Issue #30, opened 2014-10-17
libhttpd slow?


kinhunt:
Hi guys:

I have been using nodogsplash for months. All the features are great. However, when the connections go up to dozen (concurrently online, not concurrently accessing libhttpd) . The response of libhttpd will become very slow (30 seconds and more), and some times even fail to response. I've tried the ipk from repository and also the up to day source version. I've tried small box with 64mb memory and also x86 box with 2G memory. The cpu and memory usage is always low on all hardwares.  

This thread https://github.com/nodogsplash/nodogsplash/pull/9 talked about stop using multi-thread on libhttpd. Is there any thing to do with this?

Anyone have the same problem?  I will use jMeter to do some tests and update the results. Give me some heads up if you have any idea.

Thank you.

mwarning:
Threading support was added by sayuan. Does this solve the issue, @kinhunt?

kinhunt:
@mwarning @sayuan . I still cannot make it work on my devices. I will do some more tests and try to find out the patter.

PR #32, opened 2014-10-22
Turn nodogsplash back to multi-thread

sayuan:
This pull request should solve the problem of Issue #30.

I've just reverted the commit of Issue #9 without any conflict.
Then, solve the race conditions by turning static variables into local variables.


sayuan:
Well, my devices survive.

PR #39, opened 2014-12-26
stub ipv6 support

mwarning:
A first step in the direction of IPv6 support.


mwarning:
Ok, we must start IPv6 support somewhere.

florida63:
@mwarning : that's right https://github.com/nodogsplash/nodogsplash/commit/d97cc879cf59eed048606791e54eec24565cd873

I look if I find other things

Issue #46, opened 2015-01-03
replace libhttpd


mwarning:
libhttpd does not seem to be maintained anymore and lacks quality.
Any ideas for replacing it?

mwarning:
libmicrohttpd might be an option: https://www.gnu.org/software/libmicrohttpd/

lynxis:
also good candidates are libevhtp libneon. All of them are already supported by OpenWrt.

lynxis:
https://github.com/lynxis/nodogsplash/tree/microhttpd
working so far:
* mimetypes
* serving files
* templating

missing:
* auth logic

mwarning:
Great work. Let us know when it works  so we can merge it. :-)

lynxis:
I would like to merge this into a seperate branch and tag all old tickets with version "v0.9" and bugs for the "new" version with "v2". I'm not sure what we should do with v0.9. I think it's a good idea to release a version 1.0.

mwarning:
I agree, let's create release based on the current sources and create a branch for the new code for now.

nhimcoi30889:
Hi @lynxis @mwarning ,
at the moment, i can use css style for nodogsplash ( nodogsplash_0.9_beta9.9.9-6_ramips ) ? 
i tried with setup libevhtp libneon and libmicrohttpd but it dont working :(

mwarning:
@nhimcoi30889 you can use CSS just fine. There should be no problem.

bluewavenet:
... as long as any css files are in /etc/nodogsplash/htdocs

nhimcoi30889:
@mwarning : i link to css ( at /www/mycss ) , but it only show html format .
Can you help me about more detail ?
Thanks bro

nhimcoi30889:
I insert into splash.html as : 
<link href="/etc/nodogsplash/htdocs/images/bootstrap.min.css" type="text/css" rel="stylesheet"></link>
but it dont show css

bluewavenet:
That should be:
`/images/bootstrap.min.css`
not `/etc/nodogsplash/htdocs/images/bootstrap.min.css`

nhimcoi30889:
i added
`
    <link href="/images/bootstrap.min.css" type="text/css" rel="stylesheet">
    <link href="/images/font-awesome.min.css" rel="stylesheet">`
But mobile devices still dont show ssh ( it only show same type="text/html"
[...]

PR #62, opened 2015-04-02
Separation of libhttpd as separate Debian package

smoe:
Working towards a Debian package that would be acceptable for the distribution. Please give it some some critical review. It is a static library only, hence only a  -dev package is provided.


mwarning:
Does this mean a user has to download and install two packages?
This seem to be a unnecessary burden.

Anyway, there are efforts underway to replace libhttpd, as it is buggy and not maintained anymore.

smoe:
There is no extra dependency for the binary since it is only a static library that is linked against (actually, I have not changed the build script, so it is still the object files that is linked agianst if I recall/observed correctly). Hence, there is no extra dependency for the nodogsplash binary and no difference for your users. However, developers of other packages could install the libhttpd-dev package as a build dependency.

I just saw the Travis CI build failed and cannot chase that up because of a certificat issue that I fail to circumvent.

mwarning:
Hi smoe, I do not really understand the reason for packaging libhttpd. I would not recommend anyone to use it, since it is not maintained anymore and somewhat buggy.

Also, @lynxis is working towards replacing libhttpd by https://www.gnu.org/software/libmicrohttpd/.
See #46. libmicrohttpd is already in the official Debian repository. Depending on the progress, packaging libhttpd might not be necessary.

I would propose to wait a bit and see how far lynxis gets.

smoe:
Hi mwarning, I thank you for your very helpful comment. My pull request is now basically reduced to an integration of your comment with what is stated in debian/README.source , i.e., no additional package. Since the ftpmasters will otherwise request the separate offering of libhttpd, please kindly adopt / further amend that comment we (I) should then upload nodogsplash to Debian if you agree.

mwarning:
Hi smoe, I would like to merge your changes. Please smash them in one or two commits. Otherwise the changes are hard to follow.
If NDS works for you, then we could upload it to Debian.

smoe:
Please merge https://github.com/nodogsplash/nodogsplash/pull/64 instead.

PR #68, opened 2015-04-17
Added ITP to prepare for Debian upload

smoe:
It is not a blocker for the upload to the distribution (like nothing in debian/* would be), just, I would not mind to see the ITP in the regular source tree.


mwarning:
The changelog is not only for the debian package only, but the whole project (it is in debian/ thought..). But I will rework it anyway.

PR #70, opened 2015-05-08
Fixed an overflow issue in BinVoucher command line when path to BinVo…

4L3Xx:
Fixed an overflow issue in BinVoucher command line when path to BinVoucher config setting is long enough.

### DIFF
--- a/src/http.c
+++ b/src/http.c
@@ -167,7 +167,6 @@ http_nodogsplash_first_contact(request *r)
 	s_config *config;
 	const char *redir;
 	char *origurl;
-	char cmd_buff[255];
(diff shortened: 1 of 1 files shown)


mwarning:
Nice find. Thanks!

mwarning:
We need to change "sizeof(cmd_buff) - 1" to "sizeof(cmd_buff)".
Otherwise the last character will be lost.

mwarning:
Fixed in https://github.com/nodogsplash/nodogsplash/commit/6736c63e62944ecbdeb8d4aedbe7dfc7dcf9facb

PR #89, opened 2015-09-28
add domain name FirewallRules

darkkrow:
Hi!
I would need to add a domain name to the firewallrules to grant access to PreAuthenticated user because the IP address changes continuously and it would be even better if you could add a specific domain page.
In wifi dog that's already implemented; it is possible to implement that part of the code?
thanks a lot for your great work and sorry for my bad English but I'm Italian!


mwarning:
It would be possible. But somebody has to do it (as always).
Please to not create unconnected merge requests in order to ask questions.
There is a Mailinglis:t http://ml.ninux.org/mailman/listinfo/nodogsplash

PR #99, opened 2015-11-29
fix segfault in httpd_thread.c under high load

hylics:
This segfault occures if: no daemon mode AND debuglevel 7 AND high load server with *httperf --hog --server=IP --port=2050 --wsess=100,10,0.01 --rate=100 --timeout=5* AND GDB not used.

This error fixed if *httpdEndRequest(r)* in *thread_httpd(void *args)* placed after debug message which use request *r->clientAddr*


mwarning:
thanks a lot!

PR #124, opened 2016-07-20
ndsctl status durations have wrong values (thousands of days)

ansanto:
in case of enablePreAuth and binVoucher these "durations" are negative numbers (ie: durationsecs = now - client->added_time) and format_time() works with unsigned number

```
--- a/src/util.c	2016-07-20 17:10:17.000000000 +0200
+++ b/src/util.c	2016-07-20 19:11:46.927533781 +0200
@@ -364,11 +364,12 @@
 }
 
 /* Malloc's */
-char * format_time(unsigned long int secs)
+char * format_time(signed int secs)
 {
 	unsigned int days, hours, minutes, seconds;
 	char * str;
 
+	if (secs <0 ) secs =secs * -1;
 	days = secs / (24 * 60 * 60);
 	secs -= days * (24 * 60 * 60);
 	hours = secs / (60 * 60);

```


mwarning:
I do not think this is the proper solution. The input should not be negative in the first place. I will check this when I get some free time.

mwarning:
@ansanto I have committed a fix that probably solves your problem. https://github.com/nodogsplash/nodogsplash/commit/a2357e0ffd6ebad2e716fdccdf6745a2b79e1163

mwarning:
Anyway, thanks for reporting this issue!

mwarning:
The problem should be fixed by the commit mentioned.

PR #125, opened 2016-07-22
fix bug : OpenWRT cannot get mac from arp tables

chengdanhao:
I find this problem in OpenWrt Barrier Breaker 14.07

### DIFF
--- a/src/firewall.c
+++ b/src/firewall.c
@@ -89,29 +89,40 @@ unsigned int FW_MARK_MASK;             /**< @brief Iptables mask: bitwise or of 
 char *
 arp_get(const char req_ip[])
 {
-	FILE *proc;
-	char ip[INET6_ADDRSTRLEN];
-	char mac[18];
+	int type = 0;
+	int flags = 0;
+	char ip[16] = {0};
+	char mac[18] = {0};
+	char mask[16] = {0};
+	char line[128] = {0};
+	char dev[128] = {0};
 	char *reply = NULL;
+	FILE *proc_arp = NULL;
 
-	if (!(proc = fopen("/proc/net/arp", "r"))) {
+	if (!(proc_arp = fopen("/proc/net/arp", "r"))) {
 		return NULL;
 	}
 
 	/* Skip first line */
-	while (!feof(proc) && fgetc(proc) != '\n');
+	fgets(line, sizeof(line), proc_arp);
 
 	/* Find ip, copy mac in reply */
-	reply = NULL;
-	while (!feof(proc) && (fscanf(proc, " %15[0-9.] %*s %*s %17[A-Fa-f0-9:] %*s %*s", ip, mac) == 2)) {
-		if (strcmp(ip, req_ip) == 0) {
+	while (fgets(line, sizeof(line), proc_arp))
+	{
+		if (sscanf(line, "%s 0x%x 0x%x %s %s %s\n",
+				ip, &type, &flags, mac, mask, dev) < 4)
+		{
+			break;
+		}
+
+		if (0 == strcmp(ip, req_ip))
+		{
 			reply = safe_strdup(mac);
 			break;
 		}
 	}
 
-	fclose(proc);
-
+	fclose(proc_arp);
 	return reply;
 }
 


mwarning:
Thanks! I will take a look later.

src/firewall.c
```diff
@@ -89,29 +89,40 @@ unsigned int FW_MARK_MASK;             /**< @brief Iptables mask: bitwise or of
...
-	while (!feof(proc) && fgetc(proc) != '\n');
+	fgets(line, sizeof(line), proc_arp);
 
 	/* Find ip, copy mac in reply */
-	reply = NULL;
-	while (!feof(proc) && (fscanf(proc, " %15[0-9.] %*s %*s %17[A-Fa-f0-9:] %*s %*s", ip, mac) == 2)) {
-		if (strcmp(ip, req_ip) == 0) {
+	while (fgets(line, sizeof(line), proc_arp))
+	{
+		if (sscanf(line, "%s 0x%x 0x%x %s %s %s\n",
```
mwarning: We should not entirely trust /proc/net/arp.

When the output string is bigger then the buffer, the program will crash.
Also, the MAC address format needs some validation.

bluewavenet: On Barrier Breaker and Chaos Calmer /proc/net/arp is always an empty file, at least as far as I can see.  
What exactly is the issue here?
If you have the ip address and want the mac address, the place to look is /tmp/dhcp.leases.

chengdanhao: @bluewavenet 

When I run nodogsplash on my board, the syslog showed "Could not arp MAC address for 192.168.1.155". I read the source code, and find arp_get() didn't get the mac address through the client ip.

When a client connects the router, the /tmp/dhcp.leases is empty, but the /proc/net/arp has one record. This is what i see.

My firmware is built from the SDK provided by Realtek which is based on OpenWrt Barrier Breaker 14.07. They only add BSP supports for their chips.

chengdanhao: @mwarning 

Yes, you are right, my code lacks validation checks.

The arp_get function can't get MAC address through client's IP, and in fact, I am not totally understand your code, so I use a understanding way to realize my requirement.

bluewavenet: /proc/net/arp is in read only flash and is just a place holder. Arp is a function of dnsmasq in OpenWrt and the arp command will return the arp table that is held in ram by dnsmasq but this is the same data (for lan connected devices) as stored in dhcp.leases.
I did a quick test here, on Chaos Calmer, with a device with a fixed ip address and get the same syslog error as you do, yet the ip/mac is listed by the arp command. Nodogsplash does not issue a token. I have always considered this as a correct function, without looking into it, as access to the Internet through a Hotspot controller using a static ip address is not desirable. Are you using a static IP?

chengdanhao: @bluewavenet

Thanks for your reply. 

You are right. I used a staic ip in order to debugging facility.

Both my pull request and the release version get mac through the same file /proc/net/arp, but I still don't understand why does the release version can't get the device's mac with static ip.

I will read some books to figure out what the regular expression means. :-)

bluewavenet: I don't think Nodogsplash gets the mac address from /proc/net/arp. 
Here is an example of syslog as my laptop connects to the wifi, re-negotiates DHCP and has a token allocated by Nodogsplash then a few seconds later I log in. After this /proc/net/arp is still empty.

Mon Jul 25 06:54:31 2016 daemon.info hostapd: wlan0-1: STA 74:de:2b:b1:70:33 IEEE 802.11: authenticated
Mon Jul 25 06:54:31 2016 daemon.info hostapd: wlan0-1: STA 74:de:2b:b1:70:33 IEEE 802.11: associated (aid 1)
Mon Jul 25 06:54:31 2016 daemon.info dnsmasq-dhcp[1016]: DHCPREQUEST(br-lan) 10.168.10.163 74:de:2b:b1:70:33 
Mon Jul 25 06:54:31 2016 daemon.info dnsmasq-dhcp[1016]: DHCPACK(br-lan) 10.168.10.163 74:de:2b:b1:70:33 linux-8nro
Mon Jul 25 06:54:34 2016 daemon.info dnsmasq-dhcp[1016]: DHCPREQUEST(br-lan) 10.168.10.163 74:de:2b:b1:70:33 
Mon Jul 25 06:54:34 2016 daemon.info dnsmasq-dhcp[1016]: DHCPACK(br-lan) 10.168.10.163 74:de:2b:b1:70:33 linux-8nro
Mon Jul 25 06:57:49 2016 daemon.notice nodogsplash[1092]: Adding 10.168.10.163 74:de:2b:b1:70:33 token e6a826a1 to client list
Mon Jul 25 06:58:03 2016 daemon.notice nodogsplash[1092]: Authenticating 10.168.10.163 74:de:2b:b1:70:33

chengdanhao: @bluewavenet
Please look at the picture below

The redline is NodogSplash offical code, it really reads the /proc/net/arp file.

About whether /proc/net/arp is empty or not, I tried again, file in my router is not empty. Maybe our SDK have little difference.

bluewavenet: Yes I see that, but this is clearly not the only place it tries to get the mac address and this could be redundant code - I am thinking this as what you have found would fail in any hardware or version of OpenWrt I have used and I would like to have a better understanding of what is going on here.
Currently I am using Chaos Calmer and Barrier Breaker with standard releases for ar71xx and x86 and in all combinations /proc/net/arp is a read only place holder.
What is in your /proc/net/arp file? Does it relate to the ip address in question?

bluewavenet: With /proc/net/arp empty, I ran the arp command and then /proc/net/arp became a link to a file containing the same contents as the arp command output. This must be a consequence of the configuration of the squashfs file system handling, with the link to a file in ram. Interesting.
However is what you have found a bug or a consequence of using a static ip address in some way?

chengdanhao: @bluewavenet
This is my file content, I can get mac address here.


arp_get is not a redundant code, it is called by client_list_add_client。
call
code

In our practical situation, static ip is necessary, the offical way can't get the right mac, so I think this is a bug.

bluewavenet: Agreed there is something odd happening. As far as I can tell it only happens if a static ip address is used, and then not always......

mwarning:
nodogsplashs reads the IP to MAC address relationship from /proc/net/arp.
/proc/net/arp is provided by the Linux Kernel, not dnsmasq.

I expect a client using a static (IPv4) address to have an entry /proc/net/arp.
Maybe it is a timing problem and  the splash page needs to be reloaded once..

bluewavenet:
Yes timing. After a bit of testing here, with a static ip address, if the first OpenWrt sees of the client is an http request, then nodogsplash gives the "could not arp" error. Try again and it works. It seems likely that nodogsplash gets in before the arp cache is updated. If it sees a dns request first then the arp cache has updated before the http request arrives. ... For a dhcp client, all is in place well in advance.

mwarning:
For an IP address to be listed in the arp cache, there must be a communication happening between client and the computer running NDS (e.g. get a DHCP lease).

@chengdanhao: The problem cannot be fixed by this merge request, so I will close it.

PR #142, opened 2016-10-24
Debian systemd support

efernandesng:
Adds support to systemd systems (ie: debian jessie) and adds libmicrohttpd (>= 0.9.50) to package deps.

BTW: libmicrohttpd needs to be installed by tarbar or sid repository. The current version on stable repositories is lower then required version.


mwarning:
Looks good, thanks!

PR #144, opened 2016-11-10
Add initial binvoucher support

zaolin:
Hey,

WIP this includes external gateway support, check interval support and proxying of ssl requests..

Signed-off-by: Philipp Deppenwiese <zaolin@das-labor.org>

### DIFF
--- a/check-interval.rst
+++ b/check-interval.rst
@@ -1,0 +1,8 @@
+CheckInterval Option
+====================
+
+Values: 1 - (int) size
+
+This feature was introduced in previous versions of nodogsplash.
+It was used for the bin voucher support checking for invalidation
+of clients.
--- a/proxy-ssl.rst
+++ b/proxy-ssl.rst
@@ -1,0 +1,8 @@
+ProxySSL Option
+===============
+
+Values: Boolean - 1 / 0
+Default: 0
+
+The ProxySSL Option is used to enable/disable DNAT the https request to the gateway.
+This help to terminate https connections, for example on mobile phones.
--- a/src/http_microhttpd.c
+++ b/src/http_microhttpd.c
@@ -60,6 +60,7 @@ static int is_foreign_hosts(struct MHD_Connection *connection, const char *host)
 static int is_splashpage(const char *host, const char *url);
 static int get_query(struct MHD_Connection *connection, char **collect_query);
 static const char *get_redirect_url(struct MHD_Connection *connection);
+static const char *get_voucher(struct MHD_Connection *connection);
 
 static const char *lookup_mimetype(const char *filename);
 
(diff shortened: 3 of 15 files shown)


lynxis:
@zaolin can you write a documentation page about the new features - prefrable in restructured text (rst)? I'm on the way to create a docs directory with sphinx + readthedocs.

check-interval.rst
```diff
@@ -0,0 +1,8 @@
+CheckInterval Option
+====================
+
+Values: 1 - (int) size
```
mwarning: Is this in seconds?
Also (int) size is defined to be at least 32bit. Using a number here might be easier to understand to non-programmers. :-)
Maybe 2^32?

src/util.c
```diff
@@ -144,6 +149,7 @@ execute(const char cmd_line[], int quiet)
 			if(WIFEXITED(status)) {
 				debug(LOG_DEBUG, "Process PID %d exited normally, status %d", (int)rc, WEXITSTATUS(status));
 				retval = (WEXITSTATUS(status));
+				stdout = fdopen(pipes[0],"r");
```
mwarning: Make sure to close the file descriptor.

src/http_microhttpd.c
```diff
@@ -451,7 +491,30 @@ static int preauthenticated(struct MHD_Connection *connection,
 		else
 			redirect_url = get_redirect_url(connection);
 
-		if (try_to_authenticate(connection, client, host, url)) {
+		if (config->bin_voucher) {
+			if(is_alphanum(client->token)) {
+				int ret = -1;
+				char *cmd = NULL, msg[255];
```
mwarning: It is easier to understand to put each variable declarations on a line of its own.

src/util.h
```diff
@@ -31,7 +31,7 @@
 
 /** @brief Execute a shell command
  */
-int execute(const char cmd_line[], int quiet);
+int execute(const char cmd_line[], int quiet, char *msg, size_t msg_len);
```
mwarning: It might be nicer if we add default arguments to msg and msg_len. That way you do not need to add .. NULL, 0) to many execute calls.

mwarning:
@zaolin let us know when the merge request is ready to be merged.

proxy-ssl.rst
```diff
@@ -0,0 +1,8 @@
+ProxySSL Option
```
lynxis: proxyssl is a seperate feature than BinVoucher, please put this into a seperate commit.

src/http_microhttpd.c
```diff
@@ -757,9 +844,11 @@ static int show_splashpage(struct MHD_Connection *connection, t_client *client)
 	char *imagesdir = NULL;
 	char *pagesdir = NULL;
 
+	tmpl_init_templor(&templor);
+
 	memset(redirect_url_encoded, 0, sizeof(redirect_url_encoded));
 	redirect_url = get_redirect_url(connection);
-	if (redirect_url) {
+	if (strlen(redirect_url) > 0) {
```
lynxis: this is not a good idea, because get_redirect_url can return a NULL pointer. you can check both or just put the strlen() check into get_redirect_url and return NULL when strlen() == 0.

src/http_microhttpd.c
```diff
@@ -772,12 +861,10 @@ static int show_splashpage(struct MHD_Connection *connection, t_client *client)
 	safe_asprintf(&pagesdir, "/%s", config->pagesdir);
 	safe_asprintf(&imagesdir, "/%s", config->imagesdir);
 
-	tmpl_init_templor(&templor);
```
lynxis: why did you moved tmpl_init_templor()?

lynxis:
we also should not rename tok -> token becasue this would break compatibility of all templates.

zaolin:
Will split and cleanup the PR later.

drwyrm:
Any news on this? Lack of binvoucher(and a way to set per client download/upload limits) support is the only thing preventing me from using the newer code base with libmicrohttpd instead of v1.0.1.

mwarning:
@zaolin if you do not have time, let us know and someone else can step in. There is no shame in that. :-)

ram19890:
How to make BinVoucher work with 1.0.1_2, is there a documentation for setting up the BinVoucher?
If so, can someone guide me of setting up the Voucher?

PR #53, opened 2015-03-19
Automatically deauth preauthenticated users

01010000101001100:
I have noticed many preauthorised users who for unknown reasons don't go through authorisation. The major issue with it is, if they try to go through authorisation they might encounter "invalid token" error. So I automatically deauthorize them completely.

Your input, feedback and corrections to code are highly appreciated :)

PR #150, opened 2016-11-16
Bluewavenet update infoskel.html

bluewavenet:
Often, a client will connect to NDS but not authenticate for a while. When they do, very often the token will have timed out and they are served up infoskel.html.
This is neither mobile or client friendly and leaves them in limbo.
Here is a new version with a button for them to continue. Tested on Apple, Windows, Android and various desktops/laptops.

### DIFF
--- a/resources/infoskel.html
+++ b/resources/infoskel.html
@@ -1,83 +1,54 @@
+<!DOCTYPE html>
 <html>
-<!--
-	The nodogsplash server may signal an exception by serving
-	this page to the user, with variables title and content
-	intended to indicate the category and detailed message
-	corresponding to the exception.
-
-Available variables:
-	gatewayname: $gatewayname
-	version: $version
-	title: $title
-	content: $content
-	imagesdir: $imagesdir
-	pagesdir: $pagesdir
--->
 <head>
-<title>$title</title>
-<meta HTTP-EQUIV='Pragma' CONTENT='no-cache'>
+<meta http-equiv="Cache-Control" content="no-cache, no-store, must-revalidate" />
+<meta http-equiv="Pragma" content="no-cache" />
+<meta http-equiv="Expires" content="0" />
+<meta charset="utf-8" />
+<meta name="viewport" content="width=device-width, initial-scale=1.0">
+<title>Error - Something Went Wrong!</title>
 <style>
-body {
-	margin: 10px 60px 0 60px;
-	font-family : bitstream vera sans, sans-serif;
-	color: #000000;
+body {background-color:lightgrey; color:black; margin-left: 5%; margin-right: 5%;}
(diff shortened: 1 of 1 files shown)


bluewavenet:
also fixes the problem in PR #53 in a simple user friendly way.

mwarning:
This is page serves as a placeholder for the user to be replaced with a proper page.

I like to suggest the following changes:
- only minimal css for the button if any at all
- the button redirects to google.com, maybe we can use something that does not use a specific page; maybe the previous page or something like that?
- lower case html tags
- not all html, css properties on one line. It makes it hard to read.

mwarning:
That said, I can merge it now and change it afterwards. But I do not like to change merge requests a lot afterwards. :-)

bluewavenet:
A placeholder, yes, but in my opinion it also needs to "work out of the box". The main reason for this change is to make it viewable on mobile devices and give a pleasant informative experience to clients that end up on the page, usually because their token has expired. Typical Hotspot users do not seem to have much patience and the old infoskel implies something is broken when in fact, usually, nothing is wrong. 

The redirect needs to be to a valid http address on port 80. 
http://google.com is one of the most reliable and likely long-lived ones. Anyway the 99.9% of all clients (those on mobile devices) don't even see the redirect page so it hardly matters these days.

Does $redir get passed to infoskel? If so that could be used. I will try it.

I will certainly do a tidy up as suggested. The css can be simplified but mobile visibility and functionality is essential and we should deliver that on any placeholder as example. ;)

bluewavenet:
OK! $redir IS passed to infoskel. I will add this.

bluewavenet:
With the simplified css, it may well be rendered differently in different browsers, however should still be easy to read and function as designed.

bluewavenet:
Test procedure:
Connect using a smartphone etc.
When the CPD pops up, deauth the device using ndsctl in a console.
Continue on the smartphone and try to authenticate: infoskel.html should be served.
Try turning your phone through 90 degrees (with auto rotate switched on). infoskel should automatically scale to fit either way. Button should attempt to go wherever originally requested (usually determined by the CPD) and be re-captured by NodogSplash and sent back to the splash page with a fresh token.

lynxis:
@bluewavenet do you used the version 0.9/1 or version 2 for testing?

bluewavenet:
0.9_beta9.9.9 as we have this in production on CC 15.05.1 x86 and ar71xx.
Does v2 do anything different in this area?

bluewavenet:
Version 2 fails to invoke infoskel.html on an invalid token error. Instead it sends the client (confusingly and with no warning of why) back to splash.html
Tested and works as intended in version 1.0.1

bluewavenet:
infoskel-v101

PR #158, opened 2017-01-08
ndsctl: Add exception handling for the return value of send_request in the ndsctl_action

blogcin:


mwarning:
@blogcin the indentation is now a mix of tabs and spaces. Please use only tabs for this project.

Issue #159, opened 2017-01-09
v2.0.0 not work well at port 80?


jiqianghu:
Browser is long time no response when set port 80 at v2.0.0.
I use wireshark and find no http packets between browser and router.
Maybe a firewall issue?
My envirement is openwrt15.05 with libmicrohttpd library. 

(V2.0.0 with other port is ok; V0.9 is ok at port 80.)

mwarning:
Be aware that v2.0.0 has not been released yet.
Is nodogsplash running?

bluewavenet:
@jiqianghu 
On OpenWrt (or LEDE) NodogSplash will conflict with uHTTPd, also on port 80 by default.
Do you have a specific requirement to run NDS on port 80?

jiqianghu:
@mwarning :
enabled it.

@bluewavenet :
I set uhttpd another port, and set NDS at 80.
If NDS V2 support port 80, the url address will show perfect. I know v0.9 supports 80 successfully.

jiqianghu:
And another issue, to build it, openwrt1505 seems no libmicrohttpd-no-ssl, I change the make file with libmicrohttpd to comile ok. It depends other ssl libs.

bluewavenet:
@jiqianghu There may well be a bug, I have not tested this. 
With NDS running try:
http://openwrt.lan:2050/images/splash.jpg
and
http://openwrt.lan:80/images/splash.jpg
Does either of these URLs produce an image?

bluewavenet:
@jiqianghu 
In normal operation as a hotspot, https support is not needed as all devices these days do an initial request on port 80 via their built in Captive Portal Detection regardless of any "home page" that may be set in the browser. So the build will be for libmicrohttpd-no-ssl to remove unnecessary code and significantly reduce the installed size.
It would be a nice thing to have https support for some specialised applications, but would always produce a certificate error as far as I can see.

jiqianghu:
@bluewavenet 
It has response now.
But the splash.html $imagesdir is not changed.

bluewavenet:
port 80 or port 2050?
Are you changing $imagesdir for some reason?

jiqianghu:
the splash is shown, but it's image can not see. Open html source code, $imagesdir is not changed at broswser.

bluewavenet:
Did you try http://openwrt.lan:2050/images/splash.jpg ?

bluewavenet:
[...]

PR #162, opened 2017-01-11
Update http_microhttpd.c

jiqianghu:
support port 80

### DIFF
--- a/src/http_microhttpd.c
+++ b/src/http_microhttpd.c
@@ -151,8 +151,11 @@ static int is_splashpage(const char *host, const char *url)
 	} else {
 		/* hostname give - check if it's our hostname */
 
-		if (strcmp(host, our_host)) {
-			/* hostname isn't ours */
+		/* check host with port */
+		if (strcmp(host, our_host) && (80 != config->gw_port)) {
+			return 0;
+		}
+        	else if (strcmp(host, config->gw_address) && (80 == config->gw_port)) {
 			return 0;
 		}
 


mwarning:
I do not have the insight into this mechanism to know if this works correctly.
@lynxis any idea?

lynxis:
@jiqianghu please improve your commit message. The summary (first line) should contain more information than update this or that file. It should also describe why your commit was necessary.
We usually prefix the topic by the filename or the component you changed. by "microhttpd: bla".
A good description how commit messages should look like is https://who-t.blogspot.de/2009/12/on-commit-messages.html

lynxis:
@mwarning I guess we don't support nodogsplash to listen on port 80.

lynxis:
related to #159

src/http_microhttpd.c
```diff
@@ -151,8 +151,11 @@ static int is_splashpage(const char *host, const char *url)
 	} else {
 		/* hostname give - check if it's our hostname */
 
-		if (strcmp(host, our_host)) {
-			/* hostname isn't ours */
+		/* check host with port */
+		if (strcmp(host, our_host) && (80 != config->gw_port)) {
+			return 0;
+		}
+        	else if (strcmp(host, config->gw_address) && (80 == config->gw_port)) {
```
lynxis: Port 80 is more special, because both hostnames are valid:
* 192.168.1.1:80
* 192.168.1.1

mwarning: @lynxis it is not forbidden by the docs. But I think it is a good idea to not allow it until it is understood to cause no problems.

lynxis: @mwarning that's the reason, why I mentioned it. we have to support both cases in this function.

PR #163, opened 2017-01-13
ndsctl: Add exception handling for the return value of send_request i…

blogcin:
…n the ndsctl_action
---

I fix the indentation.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,14 @@
 language: c
 sudo: false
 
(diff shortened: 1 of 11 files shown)


mwarning:
@blogcin please keep in mind that it is good practice to make one pull request for each new feature. Otherwise the whole pull request might be rejected because of a single feature.

Anyway, what is this advertisment url supposed to do? :>

PR #167, opened 2017-02-15
Add binvoucher support

zaolin:
Signed-off-by: Philipp Deppenwiese <zaolin@das-labor.org>

### DIFF
--- a/src/http_microhttpd.c
+++ b/src/http_microhttpd.c
@@ -61,9 +61,47 @@ static int is_foreign_hosts(struct MHD_Connection *connection, const char *host)
 static int is_splashpage(const char *host, const char *url);
 static int get_query(struct MHD_Connection *connection, char **collect_query);
 static const char *get_redirect_url(struct MHD_Connection *connection);
-
+static const char *get_voucher(struct MHD_Connection *connection);
 static const char *lookup_mimetype(const char *filename);
 
+static int is_alphanum(const char *str)
+{
+	size_t i;
+
+	if(!str) {
+		return 0;
+	}
+
+	for (i = 0; i < strlen(str); ++i) {
+		const char c = str[i];
+		if (!((c >= '0' && c <= '9') || (c >= 'a' && c <= 'z') || (c >= 'A' && c <= 'Z')))
+			return 0;
+	}
+
+	return 1;
+}
+
+static int get_voucher_data(const char *buff, t_client *client)
+{
+	int seconds = 0;
+	int upload = 0;
+	int download = 0;
+
+	/* We require at least one value */
+	if (sscanf(buff, "%d %d %d", &seconds, &upload, &download) < 1)
+		goto err;
+	if (seconds < 1 || upload < 0 || download < 0)
+		goto err;
+
+	client->download_limit = download;
+	client->upload_limit = upload;
+	return seconds;
+
+err:
+	client->download_limit = 0;
+	client->upload_limit = 0;
+	return 0;
+}
 
 struct collect_query {
 	int i;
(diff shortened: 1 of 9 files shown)


mwarning:
@zaolin great work! Is is ready for merging? I can do some testing.

src/http_microhttpd.c
```diff
@@ -472,7 +516,35 @@ static int preauthenticated(struct MHD_Connection *connection,
 			redirect_url = get_redirect_url(connection);
 
 		if (try_to_authenticate(connection, client, host, url)) {
-			return authenticate_client(connection, ip_addr, mac, redirect_url, client);
+			if (config->bin_voucher) {
+				client->voucher = safe_strdup(get_voucher(connection));
```
mwarning: safe_strdup will exit NDs when get_voucher returns NULL

src/http_microhttpd.c
```diff
@@ -605,6 +677,18 @@ static const char *get_redirect_url(struct MHD_Connection *connection)
...
+	return query_key.value;
+}
+
+static const char *get_voucher(struct MHD_Connection *connection)
+{
+	struct collect_query_key query_key = { .key = "voucher" };
+
+	MHD_get_connection_values(connection, MHD_GET_ARGUMENT_KIND, &collect_query_key, &query_key);
+
+	if (!query_key.value)
```
mwarning: There seem to be no point to test query_key.value for NULL if we return NULL anyway in that case.

zaolin:
WIP. I need to fix the latest issues and integrate your improvements.

mwarning:
np, I am happy to see someone working on this feature. :-)

src/http_microhttpd.c
```diff
@@ -412,6 +450,12 @@ static int authenticated(struct MHD_Connection *connection,
 			return authenticate_client(connection, ip_addr, mac, redirect_url, client);
 	} else if (check_authdir_match(url, config->denydir)) {
 		auth_client_action(ip_addr, mac, AUTH_MAKE_DEAUTHENTICATED);
+		if(config->bin_voucher) {
+			char *cmd = NULL;
+
+			safe_asprintf(&cmd,"%s auth_update %s %s %d", config->bin_voucher, mac, client->voucher, time(NULL) - client->added_time);
```
lynxis: This one hurt's me, because client->voucher is filled fprintf() into cmd.

I see you sanitize the input before which is good.
Even then I like to see client->voucher ends up as as list entry of the argument list passed to execv()
I know NDS has some corpes lying around because it's using /bin/sh (even 2 times in a call - WTF).
So can you please decouple execute() a little bit more:
So that execute looks like  `execute(const char *file, const char *argv[], int quiet, char *msg, size_t msg_len)` and it's passing argument `*file` `*argv` direct execv().

If you don't like to do it, please tell me, so I can do it as requirement for your commit.

lynxis:
nice work!

zaolin:
Mhh maybe we can exchange the execute_simple implementation by my own solution via posix_spawnp:
https://github.com/zaolin/tpmcrypt/blob/devel/tools/ToolBackend.cpp

mwarning:
@zaolin ToolBackend.cpp is c++ code. The function would need to have been changed to C, of course. lynxis proposed a function signature already that looks good.
Apart from that, feel free to replace execute(..) with something better.

mwarning:
Some progress here: https://github.com/nodogsplash/nodogsplash/tree/zaolin_binvoucher

mwarning:
Looking at the code, I wonder if it is useful to allow a username and password field to be passed to the script - instead of a voucher. Using only the username or password field would replace the voucher functionality.
Would this make sense?

Cons:
- need to decode/encode username/password data as alpha-numeric chars would not be enough
pro:
- flexibility?

@lynxis any ideas?

mwarning:
@zaolin could you merge all your commits into one? I can merge this and then put my changes on top.

mwarning:
This feature is included in #260.

PR #169, opened 2017-02-19
Add disconnecting the client with arp when refreshing client list

blogcin:
nodogsplash does not automatically check client disconnects.

### DIFF
--- a/src/firewall.c
+++ b/src/firewall.c
@@ -112,6 +112,20 @@ arp_get(const char req_ip[])
 	return reply;
 }
 
+int
+is_arp_valid()
+{
+	FILE *arp_file = fopen("/proc/net/arp", "r");
+	int rc = 0;
+
+	if (arp_file) {
+		rc = 1;
+		fclose(arp_file);
+	}
+
+	return rc;
+}
+
 /** Initialize the firewall rules
  */
 int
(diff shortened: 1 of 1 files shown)


src/firewall.c
```diff
@@ -165,7 +179,18 @@ fw_refresh_client_list(void)
 			now = time(NULL);
 			last_updated = cp1->counters.last_updated;
 			added_time = cp1->added_time;
-			if (last_updated +  (config->checkinterval * config->clienttimeout) <= now) {
+
+			if (is_arp_valid() && !arp_get(ip)) {
```
mwarning: arp_get() allocates memory that must be freed by the caller. This is a memory leak. :P

mwarning:
@blogcin: this might slow down NDS a lot. For every entry, the whole arp table is iterated. /proc/arp is opened twice. Well, that can be improved.

The general idea looks fine.
@lynxis any thoughts on this one?

lynxis:
/proc/net/arp get's garbage collected when reaching certain thresholds (lowest is 128).
So I would not like to rely on this to decide if this node got disconnected or not.

lynxis:
The comment of the function is **wrong**. If we would still ping the client, we can try to look into the arp cache.
If we want to ping the client, we should ping the client using ARP rather than icmp.

blogcin:
Thanks.

PR #176, opened 2017-03-13
Add external splashpage support 

zaolin:


bluewavenet:
What does it do?
NDS can already support an external splash page via a meta redirect in splash.html

mwarning:
This code redirects to a different login page. No HTTP redirect is involved.
A drawback might be, that the redirect address has to be an IP address.

Issue #154, opened 2016-12-21
Nodogsplash not working without internet


ajaymandera:
I have successfully installed NoDogsplash captive portal on ubuntu 16.04 and it's working fine when the device is connected to LAN. Now the issue is when the device is not connected to LAN, NoDogSplash is not working and devices are not able to connect to AP or if they made a successful connection then redirection doesn't work. 
Also noticed that while not connected to LAN, NoDogsplash is eating 100% CPU.

mwarning:
What ndg version do you use?

bluewavenet:
@ajaymandera , can we assume you mean WAN rather than LAN?
NDS requires an Internet connection to work in normal circumstances as clients attaching to an AP will run their inbuilt CPD (Captive Portal Detection) and attempt to contact a set http port 80 address (determined by the operating system of the client device). The first thing the client's CPD will do after the client has received its DHCP IP address is to do a DNS query. If the Internet connection(WAN) is not available, then this will fail and the CPD will not issue its port 80 request and in turn NDS will not receive that request or serve its splash page.

In summary, what you are seeing is expected.

As for the 100% CPU, I have not seen this so cannot comment.

ajaymandera:
@mwarning  ndg version is 1.0.0

@bluewavenet no, I am talking about LAN(the incoming internet connection to my device, here PI3).
When LAN is up also WAN is up and Ndg functioning smoothly. But the problem arises when LAN is down.
Can anyone suggest me what to do in this case? I want to setup small internal network through PI with the help of ndg captive portal. 
Internet availability is not stable on PI so I want ndg to work without internet as well, is it possible? anyway I can try?

bluewavenet:
Hi @ajaymandera 
The normal terminology is the Internet feed to NDS will be its WAN and the connection to it's clients will be LAN.
It is easy to make an offline captive web site, for use in a museum for example. All you have to do is add the line:
address=/#/[ip address of your museum tour webserver]
to the config of dnsmasq.
This however will stop any access to the Internet. It should be possible to use this along with a script to detect if the Internet feed is up and modify dnsmasq config and restart it automatically.
I hope this helps ;)

bluewavenet:
@lynxis 
This will apply to version 2 as well, I think. To automate, It is more of an enhancement to Dnsmasq than NDS but would be nice to have at some stage - pretty low priority though ;)

zaolin:
@bluewavenet I will push a pull request soon which solves this issue

[...]

PR #177, opened 2017-03-13
Add dns spoofing support for offline mode and dns tunneling protection

zaolin:


lynxis:
@zaolin can you please re-base this onto master?

bluewavenet:
@zaolin 
Any progress on this. It will be VERY useful. I'm happy to help if I can.

tim-moody:
I am very interested in this PR. In my experience, it makes and installs. It also works, but I have not managed to get it to work without an internet connection. Possibly the dnsmasq options are wrong.

It is also very nearly able to be merged. Most of the conflicts are either style (// vs /* comments and spacing of log messages) or an attempt to add safe string allocation.

I would like to see this or another approach to #154 .

Justinzobel:
If this can be please re-opened and looked at it would be much appreciated. Offline mode for areas where no internet is available would be great for showing locally served pages.

bluewavenet:
This solution could be made to work but is always going to be dependent on, at least, configuration changes in whatever DNS service is in use, if even that is possible in any particular case.

The approach proposed in #328 enables NDS itself to handle all DNS requests as an inline "proxy" independent of whatever DNS service is in use.

This will give NDS the ability to autonomously detect if its Internet feed is available and react accordingly to display local html content within the standard operation of client device CPD .

For cases where an Internet connection is permanently/intentionally unavailable, modifying the config of dnsmasq to return a fixed ip address for all DNS requests works very well in conjunction with either FAS or PreAuth  being used to serve the local html without actually completing NDS authentication (thus preventing client CPD browser closure).

The only real issue preventing #328 from being implemented is time - it is somewhere near the lower end of my todo list unfortunately - ......

Justinzobel:
Is it possible to post details on a full off-line FAS or PreAuth as you mentioned? This is the result I am trying to achieve. I've got CPD working with the internet but without no luck.

bluewavenet:
@Justinzobel 
On OpenWrt, the following commands will force Offline mode:
```
uci add_list dhcp.@dnsmasq[0].address='/#/1.1.1.1'
uci commit dhcp
service dhcp restart
```
To re-enable Online mode:
```
uci del_list dhcp.@dnsmasq[0].address='/#/1.1.1.1'
uci commit dhcp
service dhcp restart
```
On other operating systems it is a similar process as long as you are using dnsmasq.

Details for using FAS and PreAuth are here:
https://nodogsplashdocs.readthedocs.io/en/stable/

Your FAS or PreAuth script should NOT authenticate to NDS as if it does, this will usually result in the client CPD browser closing (depending on vendor).

PR #211, opened 2017-11-30
Add handling calloc failure

blogcin:


bluewavenet:
@blogcin 
It looks like it makes sense but I am not seeing the full context. Can you add a description and explain the circumstances that you have found that makes this an issue?

mwarning:
@bluewavenet I think this commit is a good idea. When the system runs out of memory, allocation might fail. NDS will probably crash in  client_list_find_by_ip later when NULL is returned (as in other error cases in get_ip), but this is a step in the right direction.

bluewavenet:
@mwarning Indeed. I have not actually had a problem yet but then I have only been using LEDE on devices with 64MB ram or more.
ps It has been lonely in here of late, glad to see someone around ;-)

mwarning:
@bluewavenet I noticed, sorry about that. Life is busy and full of projects.

PR #212, opened 2017-11-30
template: support jquery syntax

blogcin:


bluewavenet:
@blogcin 
It would be useful to add a description for this please. Include why this is needed, what it fixes, what it enhances, why we need jquery support etc. Time is short as ever and any information that helps will be very useful. ;)

mwarning:
@lynxis do you know if this breaks something?

blogcin:
It was my faul.t ;(

I think that can't use jquery for template keyword parsing seems to interfere with the scalability of the splash page.

mwarning:
I don't understand the problem. Probably because I have never used jquery.
@blogcin I will close the ticket for now. Please reopen if this is still an issue for you.

Issue #216, opened 2017-12-27
Segfault when Mimetype not found


azsde:
Hello,

I am writing my own hotspot page, but I can't use FontAwesome without getting segfault when the font files are used by the program.

When I host the same page with a regular web server it works just fine.

But when I try to use it with nodogsplash, it ends up in segfault:

`[7][Wed Dec 27 17:16:57 2017][3737](src/http_microhttpd.c:252) access: GET /fonts/fontawesome-webfont.woff2
Segfault`

Here's a web page ready to be loaded (attached).

[test.zip](https://github.com/nodogsplash/nodogsplash/files/1589527/test.zip)

azsde:
This is due to font mimetypes not beeing supported.

I'm going to do a PR to fix this issue.

bluewavenet:
@azsde 
Without any other information it is rather hard to see what is going on. What hardware/os and version of NDS (must be v2)? A segfault like this is more often than not something running out of memory.

Note that the splash page in NDS is intended to be very simple, at **most** being a basic "click to enter".
I would suggest you read the Forward Authentication Service docs and follow the installation guidelines there. This will show you how to pass the required NDS data to your PHP server where the user can be authenticated before returning the token back to NDS to allow access.
https://github.com/nodogsplash/nodogsplash/tree/master/forward_authentication_service

azsde:
I've investigated and made a fix, it was due to the mimetype not being found by the **lookup_mimetype** method.

In the **mimetypes.h** the last entry was "NULL,NULL", causing a strcmp with a NULL value, which lead to a segfault.

I've removed the "NULL, NULL" entry in the mimetypes.h as it would segfault every time a mimetype wasn't found.

See my fix: https://github.com/azsde/nodogsplash/commit/51486dcb84112d95242ff86d6729f58949191ac1

bluewavenet:
PR merged.
Thanks @azsde  ;-)

PR #217, opened 2017-12-28
Fix segfault when mimetype isn't found (#216)

azsde:
The segfault was due to the mimetype not being found by the lookup_mimetype method.

In the mimetypes.h the last entry was "NULL,NULL", causing a strcmp with a NULL value, which lead to a segfault.

I've removed the "NULL, NULL" entry in the mimetypes.h as it would segfault every time a mimetype wasn't found.

I also added font related mimetypes, wasn't mandatory but it's better to have them.

### DIFF
--- a/src/http_microhttpd.c
+++ b/src/http_microhttpd.c
@@ -946,6 +946,8 @@ const char *lookup_mimetype(const char *filename)
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


bluewavenet:
Looks good - cheers!

PR #237, opened 2018-05-01
microhttpd: Use TCP Fastopen option

blogcin:
#### TCP Fast Open
An extension to speed up the opening of successive Transmission Control Protocol (TCP) connections between two endpoint.

#### libmicrohttpd support status
libmicrohttpd support [TCP Fast Open](https://en.wikipedia.org/wiki/TCP_Fast_Open) (Since 0.9.34)

nodogsplash use libmicrohttpd 0.9.51, so we can use this option to improve speed when get splash pages.

Thanks! :)

### DIFF
--- a/src/gateway.c
+++ b/src/gateway.c
@@ -255,7 +255,7 @@ main_loop(void)
 
 	/* Initializes the web server */
 	if ((webserver = MHD_start_daemon(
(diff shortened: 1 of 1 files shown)


mwarning:
Sounds good. Thanks.

PR #240, opened 2018-05-02
conf.c: Fix strncpy memory leak

blogcin:
This size make memory corruption.


mwarning:
This line would also need this change: https://github.com/nodogsplash/nodogsplash/blob/master/src/commandline.c#L78

mwarning:
Also,  config->configfile needs to initialized with 0. This would do it:
```
static s_config config = {0}
```

mwarning:
well done!

PR #260, opened 2018-07-27
binauth feature and other changes

mwarning:
This commit includes the binvoucher feature (#167) including a lot of changes.

Changes:
- binvoucher code is subsumed by the binauth setting. It can accept username and password instead of a voucher. Just the password or username alone can be used as a replacement for a voucher.
- remove setting clienttimeout and clientforceout
  - added preauth_idle_timeout  and authed_idle_timeout instead
 - now we can timeout preauthenticated and authenticated client independently
- <strike>remove unused settings passwordauth, usernameauth, username, password, remoteauthenticatoraction, enablepreauth</strike> (merged into master)
- use new execute and execute_ret methods
- struct client now has a `session_start` and `session_end` fields instead of added_time.
  - default session_end is 0 (no end)
- replace ctime as it is deprecated
- lots of code cleanups


mwarning:
@bluewavenet I hope these changes make a lot of things better. But it certainly needs testing. Any feeedback is welcome.

bluewavenet:
@mwarning
I've been tied up on another project for a couple of days. We do seem to have a great deal of common ground in terms of overall aims and we should discuss in some detail and work together. Suggestions on faster communication? I use FB messenger a great deal for this sort of thing....

mwarning:
@bluewavenet might be good idea. I do not use FB, but xmpp and irc. But there is also jitsi for voice/audio.

bluewavenet:
Is DEFAULT_REMOTE_AUTH_PORT 80 in conf.h redundant?

mwarning:
yes, it can be removed

bluewavenet:
Oops ;-) clientforcetimeout is essential for bumping out clients. Otherwise, amongst other things, a typical smartphone generating traffic every few seconds polling FB and email will stay active indefinitely.

bluewavenet:
Renaming authedidletimout back to clientidletimeout (or at lease allowing in config and aliasing) would be very good for backwards compatibility as it is the same thing, unlike old redundant options. (I have things for which this compatibility would be very convenient :-D )

mwarning:
An alias sounds good. How about clientidletimeout to set both authedidletimout and preauthidletimout?

bluewavenet:
Yes setting both from clientidletimeout. What if all three are set, or some combination? clientidletimeout overrides the other two?

bluewavenet:
clientforcetimeout is crucial. Can it be used to set the default session_end duration?

bluewavenet:
I can see how fas can call binauth as an interface for data quotas and client bumping intervals as well as all sorts of other goodies ;-)

bluewavenet:
By the way, your changes have actually fixed the issue of token expiry. NDS now issues a new token :-D

bluewavenet:
We could have session end set by clientforcetimeout and infinity if option not set, overridden by binauth.

mwarning:
@bluewavenet agreed, but the name clientforcetimeout is bugging me. clienttimeout seems to fit the meaning much better.

PR #264, opened 2018-08-01
add setting sessiontimeout

mwarning:
Reintroduces a setting clienttimeout to set the session duration of an authenticated client.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -7,7 +7,8 @@ nodogsplash (3.0.0)
       PasswordAttempts, Username, Password
   * Add binauth feature as a replacement for the old binvoucher feature
     * call a script with username/password to decide the authentication
-  * Replace ClientTimeout/ClientForceTimeout settings with PreauthIdleTimeout and AuthedIdleTimeout
+  * Replace ClientForceTimeout setting with SessionTimeout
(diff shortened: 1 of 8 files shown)


mwarning:
The code is untested and the name of the setting is not yet decided.

mwarning:
@bluewavenet what is better SessionLength or SessionDuration? Or do you have a better suggestion for a different name for ClientForceTimeout?

bluewavenet:
Personally I prefer ClientForceTimeout but I do see where you are coming from.
To have a name that unambiguously describes its function but keeping to a reasonable length is a good idea, so on that basis, I suggest:
**DefaultSessionTimeout**
It sets the default session timeout.
If not set there is no default.
Timeout rather than duration to conform with the naming of the other two idle timeout settings.

mwarning:
DefaultSessionTimeout it is then.

mwarning:
It just noticed that we already have DEFAULT_SESSION_TIMEOUT as the default for session_timeout. So instead fo renaming it to DEFAULT_DEFAULT_SESSION_TIMEOUT, I just use session_timeout the original variable name.

bluewavenet:
I'll do a compile and my own tests at some stage over the weekend :-)

PR #266, opened 2018-08-06
Rename "token" to "tok"

chrisotherwise:
The recent removal of the "token" variable stopped the "$authtarget" directive from working. This fix renames "token" to "tok" in that directive.


mwarning:
I hope most errors can be discovered before 3.0.0 is released.

PR #268, opened 2018-08-12
Forward Authentication Service (FAS) Integration.

bluewavenet:
This implements redirection to an external authentication
service in place of splash.html.

It introduces four new options:

1. fasport - if set, enables FAS redirection and the value is
the port used for access.

2. fasremoteip - if set is the ip address of the FAS, defaults
to the NDS gateway address for the case of FAS running locally to NDS.

3. faspath - the path on FAS to the login page. Default "/"

4. fas_secure_enabled - if set to "1", the client token is not
revealed and it is the responsibility of the FAS to request
this from NDS. If set to "0", the client token is passed in
clear text in the query string to the FAS.

Firewall settings are automatically configured for access to the FAS.

Binauth can also be configured at the same time to allow a local
script to be called on authentication.

### DIFF
--- a/resources/nodogsplash.conf
+++ b/resources/nodogsplash.conf
@@ -401,17 +401,16 @@ FirewallRuleSet users-to-router {
 #
 # BinAuth /bin/myauth.sh
 
-# Nodogsplash uses specific values to mark packets using iptables.
-# Parameter: FW_MARK_BLOCKED
-# Default: 0x100
+# Nodogsplash uses specific HEXADECIMAL values to mark packets used by iptables as a bitwise mask.
+# This mask can conflict with the requirements of other packages such as mwan3, sqm etc
+# Any values set here are interpreted as in hex format.
 #
-# Parameter: FW_MARK_TRUSTED
-# Default: 0x200
+# Parameter: fw_mark_authenticated
+# Default: 30000 (0011|0000|0000|0000|0000 binary)
 #
-# Parameter: FW_MARK_AUTHENTICATED
-# Default: 0x400
+# Parameter: fw_mark_trusted
+# Default: 20000 (0010|0000|0000|0000|0000 binary)
 #
-# Set FW_MARK for compatibilty with other Packages eg mwan3, sqm etc.
-fw_mark_authenticated 30000
-fw_mark_trusted 20000
-fw_mark_blocked 10000
+# Parameter: fw_mark_blocked
+# Default: 10000 (0001|0000|0000|0000|0000 binary)
+#
--- a/resources/nodogsplash.conf
+++ b/resources/nodogsplash.conf
@@ -337,6 +337,39 @@ FirewallRuleSet users-to-router {
 #
 #  DebugLevel 5
 
+# Parameter: fasport
+# Default: None
+#
+# Enable Forwarding Authentication Service (FAS)
+# If set redirection is changed from splash.html to a FAS (provided by the system administrator)
+# The value is the IP port number of the FAS
+#
+# fasport 2080
+
+# Parameter: fasremoteip
+# Default: GatewayAddress (the IP of NDS)
+#
+# If set, this is the remote ip address of the FAS.
+#
+# fasremoteip 46.32.240.41
+
+# Parameter: faspath
+# Default: /
+#
+# This is the path to the login page on the FAS.
+#
+# faspath /nodog/fas.php
+
+# Parameter: fas_secure_enabled
+# Default: 1
+#
+# If set to "1", authaction and the client token are not revealed and it is the responsibility
+# of the FAS to request the token from NDSCTL.
+# If set to "0", the client token is sent to the FAS in clear text in the query string of the
+# redirect along with authaction and redir.
+#
+# fas_secure_enabled 0
+
 # Parameter: BinAuth
 #
 # Enable BinAuth Support.
--- a/src/gateway.c
+++ b/src/gateway.c
@@ -267,6 +267,19 @@ main_loop(void)
 		httpdAddC404Content(webserver, http_nodogsplash_callback_404);
 	*/
 
+	if (config->fas_port) {
+		debug(LOG_NOTICE, "Forwarding Authentication is Enabled.\n");
+		if (config->fas_remoteip) {
+			debug(LOG_NOTICE, "FAS URL is http://%s:%u%s\n", config->fas_remoteip, config->fas_port, config->fas_path);
+		} else {
+			debug(LOG_NOTICE, "FAS URL is http://%s:%u%s\n", config->gw_address, config->fas_port, config->fas_path);
+		}
+	}
+
+	if (config->fas_secure_enabled != 1 && config->fas_port) {
+		debug(LOG_NOTICE, "Warning - Forwarding Authentication - Security is DISABLED.\n");
+	}
+
 	/* Reset the firewall (cleans it, in case we are restarting after nodogsplash crash) */
 	iptables_fw_destroy();
 
--- a/src/gateway.c
+++ b/src/gateway.c
@@ -280,6 +280,11 @@ main_loop(void)
 		debug(LOG_NOTICE, "Warning - Forwarding Authentication - Security is DISABLED.\n");
 	}
 
+	if (config->binauth) {
+		debug(LOG_NOTICE, "Binauth is Enabled.\n");
+		debug(LOG_NOTICE, "Binauth Script is %s\n", config->binauth);
+	}
+
 	/* Reset the firewall (cleans it, in case we are restarting after nodogsplash crash) */
 	iptables_fw_destroy();
 
--- a/src/http_microhttpd.c
+++ b/src/http_microhttpd.c
@@ -53,7 +53,7 @@ static int get_host_value_callback(void *cls, enum MHD_ValueKind kind, const cha
 static int serve_file(struct MHD_Connection *connection, t_client *client, const char *url);
 static int show_splashpage(struct MHD_Connection *connection, t_client *client);
 static int show_statuspage(struct MHD_Connection *connection, t_client *client);
-static int encode_and_redirect_to_splashpage(struct MHD_Connection *connection, const char *originurl);
+static int encode_and_redirect_to_splashpage(struct MHD_Connection *connection, const char *originurl, const char *querystr);
 static int redirect_to_splashpage(struct MHD_Connection *connection, t_client *client, const char *host, const char *url);
 static int send_error(struct MHD_Connection *connection, int error);
 static int send_redirect_temp(struct MHD_Connection *connection, const char *url);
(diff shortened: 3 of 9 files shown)


bluewavenet:
@mwarning

src/http_microhttpd.c
```diff
@@ -573,9 +575,11 @@ static int preauthenticated(struct MHD_Connection *connection,
  * @param originurl
  * @return
  */
-static int encode_and_redirect_to_splashpage(struct MHD_Connection *connection, const char *originurl)
+static int encode_and_redirect_to_splashpage(struct MHD_Connection *connection, const char *originurl, char *token )
 {
 	char *splashpageurl = NULL;
+	char *ip_address = NULL;
+	char *querydl = "?";
```
mwarning: better use const: `const char *querydl = "?";`

src/http_microhttpd.c
```diff
@@ -589,19 +593,37 @@ static int encode_and_redirect_to_splashpage(struct MHD_Connection *connection,
...
 
-	if (encoded[0]) {
-		safe_asprintf(&splashpageurl, "http://%s:%u/%s?redir=%s",
-			config->gw_address, config->gw_port, config->splashpage, encoded);
+	if (config->fas_port) {
+		safe_asprintf(&ip_address, "%s", config->gw_address);
+
+		if (config->fas_remoteip) {
+			safe_asprintf(&ip_address, "%s", config->fas_remoteip);
+		}
```
mwarning: This is more elegant. You also do not need to do free ip_address now.
```
ip_address = config->gw_address;
if (config->fas_remoteip) {
  ip_address = config->fas_remoteip;
}
```

src/http_microhttpd.c
```diff
@@ -589,19 +593,37 @@ static int encode_and_redirect_to_splashpage(struct MHD_Connection *connection,
...
+			safe_asprintf(&splashpageurl, "%s?clientip=%s",
+				splashpageurl, token);
+		}
 	} else {
 		safe_asprintf(&splashpageurl, "http://%s:%u/%s",
 			config->gw_address, config->gw_port, config->splashpage);
 	}
 
+	if (encoded[0]) {
+		safe_asprintf(&splashpageurl, "%s%sredir=%s", splashpageurl, querydl, encoded);
```
mwarning: You leak memory here when you use safe_asprintf multiple times on variable splashpageurl.

src/http_microhttpd.c
```diff
@@ -615,19 +627,26 @@ static int encode_and_redirect_to_splashpage(struct MHD_Connection *connection,
  */
 static int redirect_to_splashpage(struct MHD_Connection *connection, t_client *client, const char *host, const char *url)
 {
-	char *originurl = NULL;
+	const char *originurl = NULL;
```
mwarning: please remove the const here, it should not even compile

src/http_microhttpd.c
```diff
@@ -615,19 +627,26 @@ static int encode_and_redirect_to_splashpage(struct MHD_Connection *connection,
...
 	get_query(connection, &query);
 	if (!query) {
 		/* no mem */
 		return send_error(connection, 503);
 	}
-
+	if (config->fas_secure_enabled != 1) {
+		safe_asprintf(&token, "%s", client->token);
+	} else {
+		safe_asprintf(&token, "%s&gatewayname=%s", client->ip, config->gw_name);
```
mwarning: token here is either the token or some http get argument. I do not think this is a good to sneak in extra http get variables.

src/http_microhttpd.c
```diff
@@ -589,12 +593,20 @@ static int encode_and_redirect_to_splashpage(struct MHD_Connection *connection,
...
-	if (encoded[0]) {
+	if (config->fas_port) {
+		target = (config->fas_remoteip ? config->fas_remoteip : config->gw_address);
+
+		// append clientip or authaction url
+		if (config->fas_secure_enabled) {
+			safe_asprintf(&splashpageurl, "http://%s:%u/%s?clientip=%s&redir=%s",
+				target, config->fas_port, config->fas_path, token, encoded);
+		} else {
+			safe_asprintf(&splashpageurl, "http://%s:%u/%s?authaction=http://%s:%u/%s/&tok=%s&redir=%s",
```
mwarning: haven't we used authaction already for some other purpose? Or was that only an internal variable?

src/http_microhttpd.c
```diff
@@ -589,12 +593,20 @@ static int encode_and_redirect_to_splashpage(struct MHD_Connection *connection,
...
 	}
 
-	if (encoded[0]) {
+	if (config->fas_port) {
+		target = (config->fas_remoteip ? config->fas_remoteip : config->gw_address);
+
+		// append clientip or authaction url
+		if (config->fas_secure_enabled) {
+			safe_asprintf(&splashpageurl, "http://%s:%u/%s?clientip=%s&redir=%s",
+				target, config->fas_port, config->fas_path, token, encoded);
```
mwarning: clientip=<client_token>
This looks like a hack or a bug. Shouldn't it be tok=<client_token>?

src/http_microhttpd.c
bluewavenet: It would not compile without it :-D Will fix.

src/http_microhttpd.c
bluewavenet: It is a bit of a hack - actually working on it as you commented - update coming soon.

src/gateway.c
```diff
@@ -267,6 +267,19 @@ main_loop(void)
 		httpdAddC404Content(webserver, http_nodogsplash_callback_404);
 	*/
 
+	if (config->fas_port) {
+		debug(LOG_NOTICE, "Forwarding Authentication is Enabled.\n");
+		if (config->fas_remoteip) {
+		debug(LOG_NOTICE, "FAS URL is http://%s:%u%s\n", config->fas_remoteip, config->fas_port, config->fas_path);
```
mwarning: the indentation is missing here

src/http_microhttpd.c
bluewavenet: It is the exact same $authaction passed as a variable by mhd into splash.html when fas is disabled.
Fas must pass this back to nds after it has done its authentication so nds can change the client state from preauth to auth and call binauth if enabled.

src/http_microhttpd.c
bluewavenet: Same hack as mentioned above ;-)
Fix on its way.

src/http_microhttpd.c
bluewavenet: .....so nds can call binauth if enabled and change the client state from preauth to auth.

src/gateway.c
mwarning: sounds good to me.

bluewavenet:
@mwarning 
Compiled and tested on a gl-ar300m (128MB ram, 16MB flash, mips_24kc target) with Openwrt 17.01.4.

Tested with the following option combinations:
1. standard splash.html

2. splash.html with binauth

3. FAS running with lighttpd/php7 on NDS device:
fas secure off, binauth off
fas secure off, binauth on
fas secure on, binauth on
fas secure on, binauth off

4. FAS running with apache/php7 on remote multi hosted server (FAS demo):
fas secure off, binauth off
fas secure off, binauth on
fas secure on, binauth on
fas secure on, binauth off

resources/nodogsplash.conf
```diff
@@ -401,17 +401,16 @@ FirewallRuleSet users-to-router {
...
 #
-# Parameter: FW_MARK_TRUSTED
-# Default: 0x200
+# Parameter: fw_mark_authenticated
+# Default: 30000 (0011|0000|0000|0000|0000 binary)
 #
-# Parameter: FW_MARK_AUTHENTICATED
-# Default: 0x400
+# Parameter: fw_mark_trusted
+# Default: 20000 (0010|0000|0000|0000|0000 binary)
```
mwarning: This has nothing to do with fas, can you put it into a different PR?

bluewavenet: I would prefer to keep it here as the original reason for adding to config some months or so ago was to use mwan3 and sqm scripts with NDS and enable a FAS to take advantage of and  manipulate them. The same goes for binauth.
I do see your argument, but it hardly matters as soon as we merge this.

src/http_microhttpd.c
```diff
@@ -552,7 +566,7 @@ static int preauthenticated(struct MHD_Connection *connection,
 
 		if (!try_to_authenticate(connection, client, host, url)) {
 			/* user used an invalid token, redirect to splashpage but hold query "redir" intact */
-			return encode_and_redirect_to_splashpage(connection, redirect_url);
+			return encode_and_redirect_to_splashpage(connection, redirect_url, querystr);
```
mwarning: querystr will be NULL here. Won't that be a problem?

src/http_microhttpd.c
```diff
@@ -589,19 +609,26 @@ static int encode_and_redirect_to_splashpage(struct MHD_Connection *connection,
 		}
 	}
 
-	if (encoded[0]) {
+	if (config->fas_port) {
```
mwarning: Doesn't it makes sense to create the querystr here?
That is, place this block here:
```
	if (config->fas_secure_enabled != 1) {
		safe_asprintf(&querystr, "?clientip=%s&gatewayname=%s&tok=%s", client->ip, config->gw_name, client->token);
	} else {
		safe_asprintf(&querystr, "?clientip=%s&gatewayname=%s", client->ip, config->gw_name);
	}
```

mwarning:
I don't understand the whole logic yet, but the basic idea of the feature is good.
We should reorganize the commits a bit (one commit per feature and separate style changes) and merge this.

But I will need to refactor the logic in the future to make the code easier to understand and less bug prone.

src/http_microhttpd.c
bluewavenet: Not a problem and is intended. This bounces the client back to the beginning if the token is expired - just as if the CPD had triggered the splash page (be it NDS splash or FAS splash).

src/http_microhttpd.c
bluewavenet: There is an argument for that but also there is a similar argument for merging together redirect_to_splashpage and encode_and_redirect_to_splashpage.
Lots of scope for refactoring the whole of NDS as some of the logic has become a little obscured.

bluewavenet:
Re commit : some symmetry in the use of arguments
 As I was primarily concerned about fas_remoteip if fas_port is set, I put it that way round. Symmetry is good, so agree from that point of view, but in my mind the reversal points out the special case.
No problem. ;-)

mwarning:
Can you reorganize the commits? I could do it if you like.

bluewavenet:
It would be a great help if you could do it and would be much appreciated.

mwarning:
@bluewavenet hve a look. I omitted some minor code style changes. We can add them separately.
Feel free to merge.

bluewavenet:
Compiled and tested again with all combinations of FAS and Binauth.
Merging

PR #277, opened 2018-08-26
Set default gateway interface to eth0.

mwarning:
br-lan is OpenWrt specifc and uses a different configuration anyway.

### DIFF
--- a/resources/nodogsplash.conf
+++ b/resources/nodogsplash.conf
@@ -10,7 +10,7 @@
 # that is to be managed by Nodogsplash.
 # Typically br-lan for the wired and wireless lan.
 #
-GatewayInterface br-lan
+GatewayInterface eth0
 
 # FirewallRuleSet: authenticated-users
 #


mwarning:
@bluewavenet what do you think?

bluewavenet:
It will depend very much on both what the target operating system is to be and how it has been configured.
On hardware with at least 2 ethernet and and one radio it is going to be something like br-lan regardless.

On an RPi with built in radio (including the ZeroW) this will be by default wlan0.
If the RPi is using a USB WiFi dongle, the name is perhaps something related to the USB id.

The issues reported on this are I think due to the "students" clearly not following and understanding the "tutorial".
The linked tutorial:
https://pimylifeup.com/raspberry-pi-captive-portal/
Section 6 gives clear instructions.

The most likely cause of the sudden rash of similar issues is some online college/uni course having this as a project. The intention being to teach something of the inner workings. Blindly following a tutorial instead of interpreting and understanding its contents is the actual issue here.

To make this work on an RPi with a built in wireless the change would be to wlan0, rather than eth0, I think. 

However we have had br-lan as the default since pre v1.

Rather than change the default I suggest we add an extra comment eg:
"Depending on your operating system (eg Raspbian ) then you might find it useful to use the ifconfig command to see what the interface name is."
We could put this into the NDS docs as well.

PR #281, opened 2018-08-31
Add FAS redirect to status page

bluewavenet:
Redirects to FAS when status.html would otherwise be displayed.
FAS receives &status=authenticated in the query string.
It is up to FAS to interpret this and display a suitable message.

The status page is served if a client clicks "back" in their browser after authentication and tries to authenticate again when they are already authenticated.

Compiled then tested on gl.inet ar300m
 

bluewavenet:
@mwarning

PR #282, opened 2018-09-01
Add FAS and Binauth to ndsctl status

bluewavenet:
Show the configuration of both Binauth and FAS in the output of ndsctl status.

### DIFF
--- a/src/util.c
+++ b/src/util.c
@@ -438,6 +438,22 @@ ndsctl_status(FILE *fp)
 	fprintf(fp, "Managed interface: %s\n", config->gw_interface);
 	fprintf(fp, "Managed IP range: %s\n", config->gw_iprange);
 	fprintf(fp, "Server listening: %s:%d\n", config->gw_address, config->gw_port);
+
+	if (config->binauth) {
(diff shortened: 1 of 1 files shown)


bluewavenet:
@mwarning

src/util.c
```diff
@@ -438,6 +438,22 @@ ndsctl_status(FILE *fp)
...
 	fprintf(fp, "Server listening: %s:%d\n", config->gw_address, config->gw_port);
+
+	if (config->binauth) {
+		fprintf(fp, "Binauth Script: %s\n", config->binauth);
+	} else {
+		fprintf(fp, "Binauth: Disabled\n");
+	}
+
+	if (config->fas_port) {
+		fprintf(fp, "FAS: Secure=%u URL: http://%s:%u%s \n",
```
mwarning: there is an extra space before the newline

bluewavenet:
ndsctl settings
May be useful sometimes, I'll have a think about it, I'm not sure it would be worth the effort. The idea of this PR is to enable a quick check of the status of binauth and fas.

mwarning:
I just fear that `ndsctl status` gets too convoluted.

PR #283, opened 2018-09-01
initialize fas_remote with gw_address

mwarning:
This might simplify some code.


mwarning:
@bluewavenet what do you think? This should reduce some code logic.

mwarning:
The thing that is a bit odd now is that fas_remoteip is always set..

bluewavenet:
@mwarning 
I kind of had this in my mind after you mentioned it earlier. :-D
As long as fas_remoteip is always set, with your tests it will not matter.

We do need to change utils.c as well though, for the ndcstl status output.

bluewavenet:
Just compiling to give it a test...

bluewavenet:
Runs fine :-D

bluewavenet:
Some docs updates then we can go to v3.2.0 ;-)

mwarning:
nvm. I saw a few things that change the behaviour.

bluewavenet:
Re a new release:
1. ndsctl status for binauth and fas
2. Status page support for FAS
3. Iptables version check
4. Documentation updates
5. New readthedocs config
6. Initialise fas_remote with gw_address and simplify code.

PR #287, opened 2018-09-03
rename uploadbytes/downloadbytes to clientupload/clientdownload

mwarning:
Indicate that the data is client specific.


bluewavenet:
Looks good.

PR #288, opened 2018-09-03
simplify templater

mwarning:
Some code simplifications for the template engine.

### DIFF
--- a/src/http_microhttpd.c
+++ b/src/http_microhttpd.c
@@ -884,7 +884,6 @@ static int get_host_value_callback(void *cls, enum MHD_ValueKind kind, const cha
 static int show_templated_page(struct MHD_Connection *connection, t_client *client, char *page)
 {
 	struct MHD_Response *response;
-	struct templater templor;
 	s_config *config = config_get_config();
 	int ret = -1;
 	char filename[PATH_MAX];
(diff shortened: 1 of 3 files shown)


mwarning:
This also makes the code faster..

mwarning:
@bluewavenet

bluewavenet:
Looks good, indeed it should be faster. I have found the integrated FAS login page coming up noticeably quicker on my test system than the pre 3.1.0 version, as fas no longer uses this of course - now I see why.
Do you want me to compile and test?
At this rate we will keep up the release per week rate :-D

mwarning:
I haven't tested the code. But if there is something wrong with this code - we will notice it in time.
There are quite a few things that needs fixing in the code:
- NDS probably cannot deliver content for the status page like  css or image files
- the whole code in src/http_microhttpd.c is a mess that has probably a lot of missed edge cases
- traffic control is still broken
- the status page misses template variable for the number of authenticated clients
- include blocked and trusted clients in the client list (so they can be managed)

bluewavenet:
I'll test it fairly soon ;-)
It's a useful list. I think I will add it to the docs ToDo section.

 * I think NDS serves anything at any time if it is in the images directory so we probably just need to rename to something more generic eg pagedata or similar.
 * I'm not sure what management is needed for blocked and trusted clients. Both Binauth and FAS can dynamically trust or block using ndsctl. The problem arises in making them permanent ie which config file to write to. Binauth/FAS can implement degrees of blocking/trust by maintaining their own list of clients - this being much more flexible. In this way complexities are moved away from NDS itself.

bluewavenet:
In fact, I have just discovered it will serve anything from pagesdir as well, possibly since v3.0.0
So .css and .jpg files are fine!

PR #293, opened 2018-09-05
Custom tag names - html 5.2 compliance

bluewavenet:
Add hyphen to custom tag names in css
Move large comment section to after meta tag lines


mwarning:
much better

PR #302, opened 2018-09-26
Fix counter reset

### LINKED ISSUE #301: Client Upload/Download counters stop updating after session timeout and reauthentication
@mwarning 
Happens with v3x. The values given by ndsctl for a client's upload download totals at the time of session timeout, remain the same (stuck) with no further update after the client authenticates again.
At session timeout, the client changes from authenticated to preauthenticated, with the upload counters kept.

In version 2x the behaviour was (I think) to remove the client from the client list at session timeout (aka clientforcetimeout interval). Upon reconnection in v2, all the client data was reset of course.

I will have a look at the code later when I have time, but you are probably more familiar with that part than I am, so if you can take a quick look... ;-)

mwarning:
Should fix #301

### DIFF
--- a/src/auth.c
+++ b/src/auth.c
@@ -91,7 +91,7 @@ static int auth_change_state(t_client *client, const unsigned int new_state, con
 		if (new_state == FW_MARK_PREAUTHENTICATED) {
 			iptables_fw_deauthenticate(client);
 			binauth_action(client, reason);
-			client_renew_token(client);
+			client_reset(client);
 		} else if (new_state == FW_MARK_BLOCKED) {
 			return -1;
 		} else if (new_state == FW_MARK_TRUSTED) {
--- a/src/client_list.c
+++ b/src/client_list.c
@@ -49,9 +49,6 @@
 static int client_count = 0;
 static int client_id = 1;
 
-/** Time last client added */
-static unsigned long int last_client_time = 0;
-
 /** Global mutex to protect access to the client list */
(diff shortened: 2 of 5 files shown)


mwarning:
@bluewavenet can you give it a test?

bluewavenet:
Wow! I did not mean to keep you up half the night :-D
Nearly fixed.
It works with a manual ndsctl deauth, resetting the counters and issuing a new token, but sessiontimeout is ignored now.

mwarning:
Hm, what do you mean with sessiontimeout is ignored?

bluewavenet:
If sessiontimeout is set (in config), before this PR clients were deauthed automatically on the expiration of sessiontimeout. With this PR they never get deauthed automatically.

bluewavenet:
I've just noticed that when client status is preauth, ndsctl clients and json return "duration" as current time. But when authed, they show time in seconds since authentication which is as it should be.

bluewavenet:
That looks good for duration.
Looking at the code, I can't see where sessiontimeout is handled....

bluewavenet:
@mwarning 
We are all good! My fault with sessiontimeout - it is working.
The sdk I'm using had the old version of the init.d script without sessiontimeout

bluewavenet:
Github is saying Travis has not completed, but it has.
Apart from that, ready to merge.

mwarning:
Good, I couldn't see why sessiontimeout would not work.

bluewavenet:
Oh the irony! It was me that spotted the issue with init.d script - then I promptly forgot all about it.

Issue #38, opened 2014-12-26
IPv6 support


mwarning:
Hi,

NDS does not support IPv6 - but it is the future.
We could port libhttpd, bu maybe it makes sense to use a different HTTP library?

lynxis:
for both versions: The firewall handling is missing for ipv6, right?

mwarning:
yes and yes :-)

wanxewoj:
I created a pull request that fixes the IPv6 issues in `fw_iptables.c` causing nodogsplash to immediately shutdown again: https://github.com/nodogsplash/nodogsplash/pull/311

This is just another step towards IPv6 support.

wanxewoj:
Another place I've been working on is `http_microhttpd.c`. A main challenge there is that function `arp_get()` is using `/proc/net/arp` for looking up the MAC address of the requester: https://github.com/nodogsplash/nodogsplash/blob/e735416f5166df4b5f12c77af268ed9061e943bd/src/http_microhttpd.c#L231

However, this approach (via procfs) only works for IPv4, as there is no equivalent procfs entry for IPv6 neighbours. I found a generic solution via a [NETLINK_ROUTE socket](https://en.wikipedia.org/wiki/Netlink#Netlink_protocols)  that works for both IPv4 and IPv6, but it adds a new library dependency: [libnl-route](https://www.infradead.org/~tgr/libnl/doc/api/group__neigh.html#gac0f72f54c8c5f5cacf0f1aae5807ae00).

**Therefore I would like to ask first if this is acceptable?** 
For instance OpenWrt 18.06 by default only ships `libnl-tiny` which is not sufficient, as it doesn't contain `libnl-route`. NDS would require `libnl-full` in addition to `libmircohttpd` if we decide to go this direction.

mwarning:
@wanxewoj you ca use "ip neigh" to get the neighbour table. Add `-4` or `-6` if you want to have IPv4/IPv6 only.

wanxewoj:
> you ca use "ip neigh" to get the neighbour table

Sure. But then you have to parse the text output. I'm not sure if this could differ from system to system or change over time? Do we really want to consider the output of `ip neigh` as a `stable API` for NDS?

mwarning:
We currently parse text output anyway; it's not very efficient nor very stable. But it adds no dependencies and is easy to do. Also, future developers have it easier to understand the code.

wanxewoj:
[...]

PR #311, opened 2018-10-18
IPv6 firewall fixes

wanxewoj:
Another step towards [IPv6 support](https://github.com/nodogsplash/nodogsplash/issues/38). nodogsplash already has some preliminary support for IPv6 in the current `master` code base. The fixes in this pull request solve the problems in `fw_iptables.c` causing nodogsplash immediately shutting down again due to errors in setting ip6tables rules.


mwarning:
ok, looks good so far. Thanks!

PR #322, opened 2018-11-12
IPv6 address length fixes

wanxewoj:
see issue https://github.com/nodogsplash/nodogsplash/issues/38#issuecomment-437966493 (IPv6 support)


mwarning:
Thanks! :-)

PR #326, opened 2018-11-18
support authenticated user to router firewall rule.

jjm2473:
So we can control which port on router can be accessed by authenticated user now.


jjm2473:
'return to NAT table' is just in my case, I will revert it, never mind.

bluewavenet:
@jjm2473 Do you have a use case where a client gets to that rule?

jjm2473:
Yes. After 'Return to NAT table', we can redirect all DNS query to 5353, then spoof gateway ip as all A record, so we can run in offline mode.

jjm2473:
'Return to NAT table', so we have chance to handle NAT redirection ourself, not ACCEPT ALL packets;

bluewavenet:
@jjm2473 
Most operating systems in user devices have a CPD process (Captive Portal Detection).

On connecting to a network the CPD attempts to access a predefined port 80 web page. The actual URL depends on the vendor of the device and there are many of them.

First, the client device does a DNS query. If the DNS does not return with a successful response, the device will stop trying to connect. If the CPD gets the ip address resolved, it will attempt to access the page. If it does not receive the expected result it will open the CPD restricted web browser to display what it actually receives ie the splash page of the captive portal.

It is the action of trying to access the port 80 URL that preauthenticates the client.

As far as NDS is concerned if the client device is neither preauthenticated nor authenticated, all packets will be dropped.  That last rule will never be hit as far as I can tell.

If you want to make NDS work in offline mode then the DNS server must indeed spoof any queries. This must be done at the DNS server set by dhcp as it is the first thing a client device will do on connecting.

PR #327, opened 2018-11-18
Add authenticated to router firewall rule supporting.

jjm2473:
So we can control which port on router can be accessed by authenticated user now.


mwarning:
@bluewavenet I haven't tested it yet - but don't have authenticated users already access to the router?

bluewavenet:
I think users_to_router allows both preauth and auth. I can see a use for only auth to be able to access certain ports eg media server etc. (unless I have forgotten something :-D )

mwarning:
I would need to get back into understanding the iptables setup of NDS in order to review this properly.
@bluewavenet do you have an overview?

wulfy23:
Wonder if there is scope here for per device auth / pre-auth subchains....

I see scope here for nodog adapting into more of a "device classifier" than a "webtraff director"...

Seems like scalability is the number one mutually inclusive featureset.

i.e. MARK > DEVSUBCHAIN > iptables --user etc. .... just thinking aloud..... 

maybe a "MACADDRESS postauthscript = /xyz.sh" which is passed the requisite vars... spuppose it would need a $devchainname or $devmark classifier too.....

throwing some options out to extensify this thing....  much digesting to do.....

bluewavenet:
@wulfy23 

> Wonder if there is scope here for per device auth / pre-auth subchains....

Yes, the ability to do this already exists by using PreAuth (release 3.3.0 onwards), FAS (release 3.1.0 onwards) or BinAuth (release 3.0.0 onwards). Third party "Credential Verification" scripts or code can easily add device sub chains. Which method to use very much depends on what end result is desired.

In my opinion, it is much better to have NDS provide the basic Captive Portal functionality plus an API supporting development of third party applications, than it is to build in specific solutions.

> then there is the meshing with overlapping services aka sqm tags

NDS has current (and well tested) compatibility with sqm and mwan3. 
The configuration options:
fw_mark_authenticated
fw_mark_trusted
and
fw_mark_blocked
should cover just about any situation.

Proposed enhancements, API extensions or API applications are always welcome.

PR #329, opened 2018-11-26
Preauth - allow NDS to serve dynamic splash pages.

bluewavenet:
PreAuth is a means to provide a Forwarding Authentication Service (FAS) for Nodogsplash utilising NDS's built in web server.

### DIFF
--- a/docs/source/preauth.rst
+++ b/docs/source/preauth.rst
@@ -6,6 +6,8 @@ Overview
 
 PreAuth is a means to provide a Forwarding Authentication Service (FAS) for Nodogsplash (NDS) utilising NDS's built in web server.
 
+PreAuth is a means of serving dynamic html using the built in NDS web server.
+
 This means a FAS can be implemented without the resource utilisation of a separate web server, particularly useful for legacy devices with limited flash and RAM capacity.
 
 To use PreAuth, NDS FAS configuration is set to point to a virtual URL in the NDS web root instead of an external FAS server. In addition, NDS is configured with the location of the PreAuth script or program.
@@ -13,6 +15,8 @@ To use PreAuth, NDS FAS configuration is set to point to a virtual URL in the ND
 The PreAuth script can be a shell script or any other script type that an interpreter is available for (for example, PHP-cli, Python etc.).
 
 It could also be, for example, a compiled program written in C or any other language that has a compiler available for the platform.
+
+The PreAuth script or program will parse the url encoded command line (query string) passed to it and output html depending on the contents of the query string it receives from NDS. In turn, NDS will serve this html to the client device that is attempting to access the Internet.
 
 
 Using PreAuth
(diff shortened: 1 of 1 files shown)


src/http_microhttpd.c
```diff
@@ -532,8 +578,24 @@ static int preauthenticated(struct MHD_Connection *connection,
...
+	int ret;
 	s_config *config = config_get_config();
 
+	debug(LOG_DEBUG, "url: %s", url);
+	/* Check for preauthdir */
+	if (check_authdir_match(url, config->preauthdir)) {
+
+		get_query(connection, &query);
+
+		ret=show_preauthpage(connection, query);
```
mwarning: `ret = show_preauthpage(connection, query);` please add spaces :-)

forward_authentication_service/PreAuth/demo-preauth.sh
```diff
@@ -0,0 +1,73 @@
+#!/bin/sh
+query="$1"
+clientip="$(echo $query | awk -F '&' '{print $1;}' | awk -F '=' '{print $2;}')"
+gatewayname="$(echo $query | awk -F '&' '{print $2;}' | awk -F '=' '{print $2;}')"
```
mwarning: `awk -F '[&=]'` will save you an extra awk call here

mwarning:
This features seems to implement a second splash page instead of extending the current. Also, the current implementation makes it possible to pass a "rm -rf /" to the console from the web query..
I hope I can be more specific when I have more time.

forward_authentication_service/PreAuth/demo-preauth.sh
bluewavenet: That does not work unfortunately.
A typical format of the query string is:
?clientip=192.168.1.68&gatewayname=My Gateway&redir=http://fred.com&username=fred&emailaddr=fred@fredco.com
Note the leading "?" and gatewayname can have whitespace.
I am open to more efficient suggestions :-D

mwarning: Passing an URL to a script for parsing is a dangerous idea in general. Most people will get it wrong and open the device of hijacking. Better use something like this: `./script.sh 'gatewayname' 'My Gateway'`. Also, the arguments have to be sanitised (in case they contain ' or \).

bluewavenet:
The intention is to serve dynamic splash pages using MHD, not to extend the basic splash page.

mwarning:
But don't both serve the same purpose, except that one of them is generated by a script?

bluewavenet:
Yes, splash.html, FAS and FAS with Preauth all serve the same purpose to ultimately tell NDS to update iptables to allow the client access to the Internet. ie NDS authentication.

However splash.html is static. Yes we could incorporate any particular use case into the code of NDS but that is very restrictive. What happens when the next use case comes around?
Using FAS for "REAL" authentication opens the possibility of using NDS as the Captive Portal engine in just about any system.

Adding Preauth allows dynamic page generation to be served by NDS itself mitigating the need for another web server to be installed on legacy devices with limited resources.
The Preauth script could be written in any scripting language that an interpretor can be installed for, eg php-cli, python etc.
For that matter it could also be a compiled program.

mwarning:
How about we add a check for the splash page. If it is a static html file, serve it as such. If it is a script (executable), execute it and show the output as a splash page.

bluewavenet:
It is better to leave the standard NDS files (apart from config) unchanged and customise using FAS, FAS/PreAuth, FAS/Binauth, or FAS/Preauth/Binauth.
All these options can work together to enable NDS as a true Captive Portal engine, or don't switch them on for a simple splash and go solution.

bluewavenet:
Note: option preauth can be set thus activating the virtual url /nodogsplash_preauth/ even if fasremoteip and fasport are pointing to an external FAS. This is intentional as Preauth can also be used as a means of two way communication between a remote FAS and the Preauth script/program.

bluewavenet:
@mwarning 
It has now been stress tested on a large public site using a php-cli Preauth script communicating with a remote FAS, with no issues after a busy weekend.
I would like to merge after refining the documentation a bit.

src/http_microhttpd.c
```diff
@@ -759,9 +833,17 @@ static int get_query(struct MHD_Connection *connection, char **query)
 			continue;
 		}
 		strncpy(*query + j, elements[i], length - j);
+		if (i == 0) {
+			strcpy(query_str, "?");
+		} else {
+			strcat(query_str, "&");
+		}
+		strcat(query_str, *query);
```
mwarning: This (strcpy, strcat, strcat) can overflow when the input this too long.
Please use strncpy and strncat.

bluewavenet: Fixed

bluewavenet: And tested.

bluewavenet: Maybe just a wee tweak still needed......

bluewavenet: Really fixed :-D

mwarning:
@bluewavenet I cannot say that this the right approach. I just don't have the state of mind atm. to think about it. Maybe after some holiday. Feel free to merge it. But I might come up with changes to make things less complicated and haunt you. ;-)

bluewavenet:
@mwarning 
I always appreciate your input. 
I will also create additional example(s) and enhance / tidy up the documentation to go with this.

Issue #206, opened 2017-09-16
nogogsplash uci config file and init.d file


bluewavenet:
Where are they kept?
They are delivered in the LEDE/Openwrt package, but I cannot find them here and I suspect there are some ongoing issues.

bluewavenet:
Found them :-D
[https://github.com/openwrt-routing/packages](url)

bluewavenet:
Should they also be in this repository?

PR #330, opened 2018-11-27
keep openwrt package files in nds repository

mwarning:
@bluewavenet is this what you meant?
Code is sometimes easier than words. :->

### DIFF
--- a/openwrt/README.md
+++ b/openwrt/README.md
@@ -1,0 +1,68 @@
+To inlcude NoDogSplash into your OpenWRT image or to create an .ipk
+package (similar to Debians .deb files), you have to build an OpenWRT image.
+To build the firmware you need a Unix console to enter commands into.
+
+Install the dependencies of the build environment (Debian/Ubuntu):
+```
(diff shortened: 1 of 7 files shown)


bluewavenet:
#206 
Hehe :-D

mwarning:
hm, I'm okay with putting the openwrt package as it is in this PR. But not all files from the resource folder are meant for OpenWrt. That is why the files are listed in the openwrt package Makefile.

bluewavenet:
Yes, just have the openwrt-routing makefile copy restart.sh, nodogsplash (uci config), nodogsplash (init.d) and nodogsplash (uci-defaults) in exactly the same way as it currently does with splash.html, splash.css and splash.jpg.

mwarning:
The splash.* files are shared with the debian package and vanilla installation (`make install`) of NDS. So we can't move them into the `openwrt/nodogsplash/files/`.

mwarning:
ok, since we agree on this part so far.

PR #331, opened 2018-11-28
Copy OpenWrt specific files

bluewavenet:
Makefile: Openwrt specific files copied from PKG_SOURCE

This makefile should then also be used in openwrt-packages.
Then for an official release of OpenWrt package only the PKG_VERSION, PKG_RELEASE and PKG_HASH need to be updated with no access to the /nodogsplash/nodogsplash/ repository being required by a maintainer.

### DIFF
--- a/openwrt/nodogsplash/Makefile
+++ b/openwrt/nodogsplash/Makefile
@@ -41,7 +41,6 @@ define Package/nodogsplash/description
 endef
 
 define Package/nodogsplash/install
-	$(CP) ./files/* $(1)/
 
 	$(INSTALL_DIR) $(1)/usr/bin
 	$(INSTALL_BIN) $(PKG_BUILD_DIR)/nodogsplash $(1)/usr/bin/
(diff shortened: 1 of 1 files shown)


bluewavenet:
@mwarning
Code is sometimes easier than words. :-D

mwarning:
I would prefer to have the OpenWrt specific stuff limited to the openwrt folder the way it is right now. Otherwise the package maintenance will be tied to this repository - something I like to avoid.

The *.html files etc. are not packet specific and as such are located in the nds repository.

bluewavenet:
Keeping OpenWrt specific stuff limited to the openwrt folder is exactly what I am saying.
Package maintenance will be made simpler and less prone to errors by using this makefile.

bluewavenet:
I do see your point of view though.
How can we do this? How to make releasing on Openwrt simpler and less error prone?
Back to the fork of openwrt-routing idea?

PR #334, opened 2018-12-17
WIP: Per-user traffic quota (not bandwith, but total transfer limit)

jsphpl:
I am starting to work on the implementation of total transfer quota on a per-user basis. This obviously is by no means anywhere near ready to be merged. I am opening this pull request in order to ask a few questions during development.

First question: what would be the right thing to do if the user exceeds their quota? Right now, i am calling `iptables_fw_deauthenticate(client)`, just because it is the only function that accepts a `t_client` argument and sounds like it would lock them out. I clearly do not have an understanding of how everything works, yet. So a small hint would be appreciated, in order to get me going into the right direction.

I am also thankful for any other hints supporting my current goal ;)

Thanks a lot in advance!

### DIFF
--- a/src/fw_iptables.c
+++ b/src/fw_iptables.c
@@ -382,7 +382,7 @@ iptables_fw_init(void)
 	LOCK_CONFIG();
 	config = config_get_config();
 	gw_interface = safe_strdup(config->gw_interface); /* must free */
-	
+
 	/* ip4 vs ip6 differences */
 	const char *ICMP_TYPE;
 	if (config->ip6) {
@@ -393,7 +393,7 @@ iptables_fw_init(void)
 		gw_ip = safe_strdup(config->gw_ip);    /* must free */
 		ICMP_TYPE = "icmp";
 	}
-	
+
 	gw_address = safe_strdup(config->gw_address);    /* must free */
 	gw_iprange = safe_strdup(config->gw_iprange);    /* must free */
 	gw_port = config->gw_port;
@@ -497,7 +497,7 @@ iptables_fw_init(void)
 	 * Set up nat table chains and rules (ip4 only)
 	 *
 	 */
-	 
+
 	if (!config->ip6) {
 		/* Create new chains in nat table */
 		rc |= iptables_do_command("-t nat -N " CHAIN_OUTGOING);
@@ -1065,6 +1065,7 @@ iptables_fw_counters_update(void)
 				if (p1->counters.incoming < counter) {
 					p1->counters.incoming = counter;
 					debug(LOG_DEBUG, "%s - Updated counter.incoming to %llu bytes", ip, counter);
+					iptables_block_if_counter_exceeded(p1);
 				}
 			} else {
 				debug(LOG_WARNING, "Could not find %s in client list", ip);
@@ -1075,4 +1076,14 @@ iptables_fw_counters_update(void)
 	pclose(output);
 
 	return 0;
+}
+
+void
+iptables_block_if_counter_exceeded(t_client *client)
+{
+	unsigned long long int incoming_limit = 100000;
+	if (client->counters.incoming > incoming_limit) {
+		debug(LOG_DEBUG, "Client has exceeded maximum limit. Deauthenticating.");
+		iptables_fw_deauthenticate(client);
+	}
 }
--- a/src/fw_iptables.h
+++ b/src/fw_iptables.h
@@ -90,4 +90,6 @@ int iptables_unallow_mac(const char mac[]);
 int iptables_trust_mac(const char mac[]);
 int iptables_untrust_mac(const char mac[]);
 
+void iptables_block_if_counter_exceeded(t_client *client);
+
 #endif /* _IPTABLES_H_ */


mwarning:
Hi

I have two snippets that might help. Let me drop them here for you to recycle (if useful), since I do not have time right now.

```
src/auth.c
@@ -185,6 +185,16 @@ fw_refresh_client_list(void)
                                cp1->counters.incoming / 1000, cp1->counters.outgoing / 1000);
 
                        auth_change_state(cp1, FW_MARK_PREAUTHENTICATED, "idle_deauth");
+               // hm, really logout? or just activate minimal traffic mode?
+               } else if (conn_state == FW_MARK_AUTHENTICATED
+                               && cp1->traffic_limit
+                               && ((cp1->counters.incoming + cp1->counters.outgoing) / 1000) > cp1->traffic_limit) {
+
+                       debug(LOG_NOTICE, "Logout authenticated user because download limit was reached: in: %llukB, out: %llukB, limit: %llukB",
+                               cp1->ip, cp1->mac, cp1->counters.incoming / 1000, cp1->counters.outgoing / 1000, cp1->traffic_limit);
+
+                       // or activate traffic limit code, like 8kb/s
+                       auth_change_state(cp1, FW_MARK_PREAUTHENTICATED, "traffic_deauth");
```

```
src/http_microhttpd.c
@@ -68,16 +68,17 @@ static const char *lookup_mimetype(const char *filename);
 
 /* Get client settings from binauth */
 static int do_binauth(struct MHD_Connection *connection, const char *binauth, t_client *client,
-       int *seconds_ret, int *upload_ret, int *download_ret)
+       int *seconds_ret, int *upload_ret, int *download_ret, int *traffic_ret)
 {
        char username_enc[64] = {0};
        char password_enc[64] = {0};
        const char *username;
        const char *password;
        char msg[255] = {0};
-       int seconds;
-       int upload;
-       int download;
+       int seconds = 0; // session duration
+       int upload = 0; // rate in kB/s
+       int download = 0; // rate in kB/s
+       int traffic = 0; // volume in kB
        int rc;
 
        username = MHD_lookup_connection_value(connection, MHD_GET_ARGUMENT_KIND, "username");
@@ -89,17 +90,19 @@ static int do_binauth(struct MHD_Connection *connection, const char *binauth, t_
                return -1;
        }
 
-       rc = execute_ret(msg, sizeof(msg) - 1, "%s auth_client %s '%s' '%s'",
-               binauth, client->mac, username_enc, password_enc);
+       rc = execute_ret(msg, sizeof(msg) - 1, "%s auth_client %s '%s' '%s' %d",
+               binauth, client->mac, username_enc, password_enc, client->traffic_limit);
 
        if (rc != 0) {
                return -1;
        }
 
-       rc = sscanf(msg, "%d %d %d", &seconds, &upload, &download);
+       rc = sscanf(msg, "%d %d %d %d", &seconds, &upload, &download, &traffic);
 
        // store assigned parameters
        switch (rc) {
+               case 4:
+                       *traffic_ret = MAX(traffic, 0);
                case 3:
```

bluewavenet:
@jsphpl 
Hooks to do just this are already in place, directly via Binauth. This allows you to enforce either a preset quota for everyone or if you implement your own user database, individual quotas. 
It can also be done quite simply using FAS and/or Preauth if you want the preset quota for everyone.

Note that Binauth is a post authentication hook, whilst Preauth is a pre authentication hook.
The snippets provided by @mwarning are from the Binauth code.

Using a combination of both it is fairly simple to implement a means of checking a user database for individual quotas or quotas per time interval and give a dynamic response page to the user as they log in, for example, "Yes you can log in and have xxMB left", or "Sorry you have used your quota for today" etc., with Preauth doing the preauthentication user/password/quota checking, and Binauth setting the individual quotas. The quota set for a particular user can even be [Daily Quota] minus [amount used in last 24 hours].
Have a play about with BinAuth and Preauth to get an understanding how it works and you will see what I mean.
Documentation at:
https://nodogsplashdocs.readthedocs.io/en/latest/

The advantage of doing it this way involves no further modifications to the NDS code, thus keeping it "un-bloated" and simple to maintain. (I have in fact implemented a quota system this way for private projects so am 100% confident it works).

In my opinion, what you are suggesting is very worth while implementing, but should be done using Preauth/Binauth. It can be a C program, called by both Preauth and Binauth that maintains its own built in user database. It could of course also be written in anything available for the target platform eg php, python, shell script etc.. Writing it in C as a separate program will make it very safe for the inexperienced, and available in the NDS package.

bluewavenet:
@jsphpl 
Binauth does not directly support quotas, but a combination of it with Preauth can easily be made to implement quotas.
The key is using ndsctl to get the current data usage for a client.

bluewavenet:
@jsphpl 
In a simple stand alone system (ie no external FAS), it is most simply done just with Preauth.
A Preauth program would check its database and respond to the client appropriately. 
In my implementation I have a simple shell script running in the background that every 2 minutes checks authenticated client data totals and deauths any clients that have exceeded their quota.
This background process is started by my Preauth program on its first call after starting NDS.

mwarning:
Currently, NDS only reports the traffic to the binauth script (if configured) on user logout/timeout. But NDS does not take any action when a quota is reached.
So that would be a nice feature.

jsphpl:
Btw for anyone stumbling across this thread: here's an external per-user quota implementaion using preauth and written in python: https://github.com/jsphpl/nds-quota

Still undocumented, sorry. Very basic instructions in https://github.com/jsphpl/nds-quota/issues/1

bluewavenet:
@jsphpl 
Nice to see a python preauth ;-)

PR #343, opened 2019-03-19
for upstream

lynxis:


openwrt/nodogsplash/files/etc/init.d/nodogsplash
```diff
@@ -172,7 +177,10 @@ create_instance() {
   config_get_bool val "$cfg" enabled 0
   [ $val -gt 0 ] || return 0
 
-  generate_uci_config "$cfg"
+  if ! generate_uci_config "$cfg" ; then
+      echo "Can not generate uci config. Will not start instance $cfg." >&2
+      return
```
mwarning: shouldn't this be `return 1`?

mwarning:
It looks good enough. We can fix minor things afterwards if necessary.

PR #344, opened 2019-03-19
remove dead code for removed network setting

mwarning:
@bluewavenet can you take a look. In case I missed something. The network setting has been removed und essentially merge with gatewayinterface.

### DIFF
--- a/openwrt/nodogsplash/files/etc/config/nodogsplash
+++ b/openwrt/nodogsplash/files/etc/config/nodogsplash
@@ -109,7 +109,7 @@ config nodogsplash
   #option preauth '/bin/mypreauth.sh'
 
   # Your router may have several interfaces, and you
-  # probably want to keep them private from the network/gatewayinterface.
+  # probably want to keep them private from the gatewayinterface.
   # If so, you should block the entire subnets on those interfaces, e.g.:
   #list authenticated_users 'block to 192.168.0.0/16'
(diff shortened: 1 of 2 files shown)


bluewavenet:
Looks good. Default gatewayinterface is br-lan in uci config, but is there a case for setting this in conf.h?

mwarning:
I do not understand the question. gatewayinterface is stored in setting gw_interface of conf.h.

mwarning:
I do not understand the question. gatewayinterface is stored in setting gw_interface of conf.h.

bluewavenet:
Ah, yes sorry, I meant to say set a default in conf.h
ie:
`#define DEFAULT_GATEWAYINTERFACE "br-lan"`
and use this to fill gw_interface if not supplied in config file.

mwarning:
br-lan is openwrt specific. That default should stay in /etc/config/nodogsplash.

mwarning:
Anyway, that is a different discussion anyway.

bluewavenet:
If not set in config it just errors and fails, always has, so not a problem.

I did not think br-lan was openwrt specific, rather, a very common choice for a bridge interface name on a lan. It is also the default in the nodogsplash.conf file used for other Linux distributions eg in my experience, Suse Enterprise and OpenSuse where it is the default bridge interface name as far as I remember.
No matter though.

PR #347, opened 2019-04-03
Prevent access to the basic splash page (splash.html) when FAS is enabled.

bluewavenet:
When FAS is enabled, returns "Error 403 Forbidden" if http://gw_address:gw_port is accessed directly (rather than being triggered by CPD or trying to access a port 80 web site). 


bluewavenet:
Compiled and tested on OpenWrt mips_24kc target.

PR #360, opened 2019-04-22
 Validate Fasremoteip and Fasport 

bluewavenet:
1. fasremoteip must be a dot format ipv4 address. The config value is now validated and NDS terminates with a console error message if validation fails.
 2. Version 3.3.0 introduced improvements towards usable IPv6 support. This however induces a "too many redirects error" in device CPD if port 80 is used for FAS in the case of FAS running locally on the NDS device. Fasport is now checked for port 80 and local FAS, and NDS terminates with a console error message if these conditions are found.
 3. The config file has been updated to reflect points 1 and 2.
 4. Documentation has been updated to reflect points 1 and 2.

Compile tested on OpenWrt SDK, ath79 platform.

Run tested on gl-inet ar300m-16

### DIFF
--- a/debian/doc/nodogsplash.1
+++ b/debian/doc/nodogsplash.1
@@ -1,6 +1,6 @@
 .\" Man page generated from reStructuredText.
 .
-.TH "NODOGSPLASH" "1" "April 05, 2019" "3.3.1-beta" "nodogsplash"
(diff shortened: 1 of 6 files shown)


bluewavenet:
C99 mode? Pfft : -D

PR #369, opened 2019-06-14
openwrt: initscript: add missing macmechanism in the config file

lynxis:
The init script could parse the macmechanism and the mac lists, but it forgot to
add the macmechanism to the configuration itself


bluewavenet:
@mwarning 
Agreed, but I also have the feeling something else is missing from there but have not had the time to look.
I also have a couple of other things planned that would be useful and some documentation updates to do.

I am finishing a project this weekend so can do all this early next week with luck if a new release can wait a few days.

PR #372, opened 2019-06-19
use elegant check for valid ip addresses

mwarning:
A less error prone method to check for a valid ip address.

### DIFF
--- a/src/util.c
+++ b/src/util.c
@@ -373,6 +373,14 @@ get_uptime_string(char buf[64])
 	return format_duration(started_time, time(NULL), buf);
 }
 
+int is_addr(const char* addr) {
+	struct sockaddr_in sa;
+	struct sockaddr_in6 sa6;
+
+	return (inet_pton(AF_INET, addr, &sa.sin_addr) == 1) ||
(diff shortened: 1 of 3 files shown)


mwarning:
@bluewavenet let me know if this is ok for you, since you wrote the previous code.

bluewavenet:
@mwarning 
I'm happy if it works, presumably it does! :-D
I spent far too much time trying to get inet_pton to work, then gave up.

src/util.c
```diff
@@ -373,6 +373,14 @@ get_uptime_string(char buf[64])
 	return format_duration(started_time, time(NULL), buf);
 }
 
+int is_addr(const char* addr) {
+	struct sockaddr_in sa;
+	struct sockaddr_in6 sa6;
+
+	return (inet_pton(AF_INET, addr, &sa.sin_addr) == 1) ||
+		(inet_pton(AF_INET6, addr, &sa6.sin6_addr) == 1);
```
mwarning: hm, the || should have gone into the next line. Well, that's for the next code cleanup then..

PR #377, opened 2019-06-27
Allow FQDN for local gateway URLs

pallas:


bluewavenet:
@pallas 
Could you add a description to the pr and not just the commits, thanks

bluewavenet:
> This is useful, for instance, if you want to replace an IP address with a
> local gateway domain name.

Sorry, I must be I missing something, what is the use of this? NDS must run on a router and for the captive portal to function it must also be the local network gateway and the gateway address is automatically detected. The ability to provide this address in the config is provided only to cater for the case where in some operating system the automatic detection of the router's LAN ip address fails or the router has multiple ip addresses on the same LAN.

bluewavenet:
> It may be worth adding a similar option to FAS, so that we can allow the IP
> but redirect to the FQDN.

Yes, that is worth looking at for the case of the FAS server being hosted on a shared server.

> This would more easily enable the remote FAS to be https.

Unfortunately not. NDS currently forces an http redirect. If it did not and allowed an https redirect, as soon as the FAS attempted to authenticate back to NDS the client CDP or browser will immediately issue a security warning something like "link to insecure web page".

bluewavenet:
>  authdir: modify matching to check for /s, only require lead

> We've seen some 3rd party splash that doesn't honor the cb exactly.

"cb" ?

> We've seen some 3rd party splash that doesn't honor the cb exactly.
> Specifically, it parses it and then reforms it, cutting off the trailing
> slash.

Do you mean a third party splash.html or a third party FAS? If so that third party code should be fixed rather than modify NDS to cater for someone's error.
authdir is by default nodogsplash_auth so a typical call to NDS for authentication will be something like:
http://192.168.10.1:2050/nodogsplash_auth/?tok=d697a226&redir=http://detectportal.firefox.com/success.txt

PR #378, opened 2019-07-01
remove unused pagesdir and imagesdir

mwarning:
Both variables are not used anymore.

Iirc, the original purpose was to have explicit subfolders so to avoid path name clashes when the intercepted page was loaded. Otherwise, splash page content might be shown in the intercepted page.

### DIFF
--- a/openwrt/nodogsplash/files/etc/init.d/nodogsplash
+++ b/openwrt/nodogsplash/files/etc/init.d/nodogsplash
@@ -141,7 +141,7 @@ generate_uci_config() {
 
   for option in preauth binauth fasport fasremoteip faspath fas_secure_enabled \
     daemon debuglevel maxclients gatewayname gatewayinterface gatewayiprange \
-    gatewayaddress gatewayport webroot splashpage statuspage imagesdir pagesdir \
+    gatewayaddress gatewayport webroot splashpage statuspage \
     redirecturl sessiontimeout preauthidletimeout authidletimeout checkinterval \
     setmss mssvalue trafficcontrol downloadlimit uploadlimit \
     syslogfacility ndsctlsocket fw_mark_authenticated \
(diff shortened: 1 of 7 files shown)


mwarning:
Since this might breaks old setups, a major release version change would be required.

bluewavenet:
Ah! ftpmirror.gnu.org is currently down - hence the Travis failure...

bluewavenet:
@mwarning 
I remember trying changing these settings just to see if they worked, but apart from that I have never needed them so agree with their removal.
Should we also remove **webroot** and **splashpage** as config options?

I have a couple of significant enhancements to FAS almost ready for PR so If I can get that finished it would be a good time for the major release. I assume you mean v4.0.0 ....

mwarning:
`webroot` and `splashpage` are used and important options to have. `webroot` defaults to `/etc/nodogsplash/htdocs`, `splashpage` defaults to `splash.html`.

mwarning:
Yes, I mean 4.0.0 (v4.0.0 is the git tag).

bluewavenet:
My thoughts were if we don't need to change pagesdir or imagesdir, we won't need to change webroot or splashpage either. 

My preference would be to remove all four as the default for these is all that is really needed, particularly with CPD of modern devices making dynamic "splash" pages essential for anything other than the most basic "Tap/Click Continue" configuration.
Just my opinion :-D

mwarning:
Setting pagesdir or imagesdir has no effect on the behavior of the program. Except for that it is substituted in the templates.

bluewavenet:
I haven't checked the code but are you saying pagesdir and imagesdir have no effect (apart from template variables)? They used to work a long time ago....

bluewavenet:
@mwarning 
I'm ok with merging this regardless of my comments :-D

mwarning:
The usage of those variables was removed as part of the switch to libmicrohttpd. Since then the serve no purpose anymore.

PR #381, opened 2019-07-07
Add fasremotefqdn, faskey, bump to v4.0.0

bluewavenet:
From the Changelog:
  * Introduce aes encryption of the query string passed to remote FAS, allowing authdir and client token to be transferred securely. Uses php-cli and php-openssl. These are required if encryption is enabled but are not dependencies [bluewavenet]
  * Introduce fasremotefqdn, specifying the FQDN of the remote FAS. This facilitates simplified support for FAS operation on shared hosting systems [bluewavenet]
  * Add a FAS php script supporting aes encrypted query string sent from NDS [bluewavenet]
  * Numerous Documentation updates [bluewavenet]

Compiled using OpenWrt SDK-snapshot for target ath79
Package tested on GL-AR300M-16 and GL-AR750 hardware.


bluewavenet:
@mwarning

mwarning:
hm, this is a lot.

docs/source/conf.py
```diff
@@ -60,9 +60,9 @@
 # built documents.
 #
 # The short X.Y version.
-version = '3.3.3-beta'
+version = '4.0.0'
 # The full version, including alpha/beta/rc tags.
-release = '3.3.3-beta'
+release = '4.0.0-1'
```
mwarning: The -N suffixes are for package releases, not the source code.

openwrt/nodogsplash/Makefile
```diff
@@ -7,8 +7,8 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=nodogsplash
 PKG_FIXUP:=autoreconf
-PKG_VERSION:=3.3.3
-PKG_RELEASE:=1-beta
+PKG_VERSION:=4.0.0-beta
```
mwarning: 4.0.0-beta?

docs/source/conf.py
bluewavenet: Well yes but conf.py has been like this for as long as I know. Also the intention is to merge this as beta then update to release.... I can fix this now if you think it would be better :-D

openwrt/nodogsplash/Makefile
bluewavenet: Yes - see previous comment. I have been compiling the package and testing and it is useful to see which test version you are at. Can fix these now if you like.

bluewavenet:
I've now completed a 7 day test in a live environment with no issues.
I'll merge then check the documentation updates ready for a release.

PR #392, opened 2019-07-23
make debuglevel platform independent

mwarning:
Make the debug level  platform independent and less confusing.
Now the levels are 0, 1, 2. We can add more if needed.
The default log level is 1.

### DIFF
--- a/src/commandline.c
+++ b/src/commandline.c
@@ -48,12 +48,12 @@ usage(void)
 		"\n"
 		"  -c <path>   Use configuration file\n"
 		"  -f          Run in foreground\n"
-		"  -d <level>  Debug level (0-9)\n"
+		"  -d <level>  Debug level (%d-%d)\n"
 		"  -s          Log to syslog\n"
 		"  -w <path>   Ndsctl socket path\n"
 		"  -h          Print this help\n"
 		"  -v          Print version\n"
-		"\n"
+		"\n", DEBUGLEVEL_MIN, DEBUGLEVEL_MAX
 	);
 }
 
@@ -92,8 +92,9 @@ void parse_commandline(int argc, char **argv)
 			break;
 
 		case 'd':
-			if (optarg) {
-				set_log_level(atoi(optarg));
+			if (set_debuglevel(optarg)) {
+				printf("Could not set debuglevel to %d\n", atoi(optarg));
+				exit(1);
 			}
 			break;
 
--- a/src/conf.c
+++ b/src/conf.c
@@ -735,11 +735,14 @@ config_read(const char *filename)
 			}
 			break;
 		case oDebugLevel:
-			if (sscanf(p1, "%d", &config.debuglevel) < 1 || config.debuglevel < LOG_EMERG || config.debuglevel > LOG_DEBUG) {
-				debug(LOG_ERR, "Bad arg %s to option %s on line %d in %s. Valid debuglevel %d..%d",
-					p1, s, linenum, filename, LOG_EMERG, LOG_DEBUG);
+			if (sscanf(p1, "%d", &config.debuglevel) < 1 || config.debuglevel < DEBUGLEVEL_MIN) {
+				debug(LOG_ERR, "Bad arg %s to option %s on line %d in %s. Valid levels are %d...%d.",
+					p1, s, linenum, filename, DEBUGLEVEL_MIN, DEBUGLEVEL_MAX);
 				debug(LOG_ERR, "Exiting...");
 				exit(1);
+			} else if (config.debuglevel > DEBUGLEVEL_MAX) {
+				config.debuglevel = DEBUGLEVEL_MAX;
+				debug(LOG_WARNING, "Invalid debug level. Set to maximum.");
 			}
 			break;
 		case oMaxClients:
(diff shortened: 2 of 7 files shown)


bluewavenet:
@mwarning 
This makes much more sense ;-)
If I am reading correctly, the mapping of levels, old to new is:
0 => 0
5 => 1
7 => 2

The same checks need to be done in ndsctl (in ndsctl_thread.c ?). 
ndsctl currently uses the keyword loglevel rather then debuglevel, should we change this?
The config files will need to be updated, as will the documentation.

mwarning:
Yes, loglevel is better. But debuglevel still needs to be supported for some time.

bluewavenet:
Actually I was thinking the other way around ie change ndsctl to use debuglevel :-D
If you can add the ndsctl checks and merge, I will do the config files and documentation if you like...

mwarning:
The code is not tested yet.
Debug levels:
0: Only display errors
1. Display errors, warnings, basic information text
2. Output everything

Default debuglevel is 1.

@bluewavenet debuglevel is a bit of an odd name. This is more about verbosity.

bluewavenet:
@mwarning 
I'm all set up here and can compile and test whenever you are ready.

mwarning:
@bluewavenet sure, give a spin.

bluewavenet:
@mwarning 
Mmm. There is something wrong with the logic. In the old scheme we had (I think):
LOG_EMERG displayed at level 0, level 5, and level 7
LOG_ERR displayed at level 0, level 5, and level 7

LOG_WARNING displayed at level 5 and level  7
LOG_NOTICE displayed at level 5 and level 7

LOG_INFO displayed at level 7
LOG_DEBUG displayed at level 7

In the new scheme it looks like:
LOG_EMERG displayed at level 0
LOG_ERR displayed at level 0

LOG_WARNING displayed at level  1
LOG_INFO displayed at level 1

LOG_NOTICE displayed at level 2
LOG_DEBUGdisplayed at level 2

The result is a bit of a mess :-D

ndsctl debuglevel _n_
returns
Unknown command: debuglevel

and 
ndsctl loglevel _n_
returns
ndsctl: Error: nodogsplash sent an abnormal reply.

mwarning:
@bluewavenet I've fixed a few things, but the debuglevel was working for me.

Anyway, we should rename debuglevel to loglevel or something different. The problem is that we break the meaning of debuglevel. We need to introduce a new setting and deprecate the old.

mwarning:
I've added another debug level. Because it is useful to differentiate between verbose and debug messages.

bluewavenet:
@mwarning 
Noooo! It is now completely broken.
On upgrading to this from a working binary immediately prior to this merge, all I now get is:

> Sun Jul 28 07:33:24 2019 daemon.info procd: Instance nodogsplash::cfg015847 s in a crash loop 6 crashes, 0 seconds since last crash

There are no other entries in logread about NDS.
It is the same for   option debuglevel  0, 1, 2, 3 and not set.
So of course I don't get any indication of what is wrong either.
The binary I am compiling using snapshot SDK from current master is mips_24kc.

bluewavenet:
@mwarning 
I think we need something much more flexible, yet simple to understand.
ie:
debuglevel a two digit hex number.

The following reporting levels:
SILENT 0x00 00000000
EMERG 0x01 00000001
ERR 0x02 00000010
WARNING 0x04 00000100
NOTICE 0x08 00001000
INFO 0x10 00010000
DEBUG 0x20 00100000
RESERVED1 0x40 01000000
RESERVED2 0x80 10000000

Then we can set the default to 0x0F, giving EMERG+ERR+WARNING+NOTICE, the equivalent of the old way's default of '5'.

We could simply set 'Fully Verbose' as 0xFF that includes the two additional levels RESERVED1 and RESERVED2 that could be used for future expansion.

The new completely silent mode, SILENT 0x00, might be desired in some circumstances.

This will give us the ultimate in flexibility and be easy for people to understand (and easy to document!).

mwarning:
I will take another look.
Anyway, hex number or decimal number make no difference. Maybe you want debug levels like 'quiet', ''default', 'verbose' or 'debug'?

As for RESERVED1, it is probably called reserved for a reason. The whole point of the commit was not to rely on the definition of the syslog level implementation.

bluewavenet:
@mwarning 
Currently master is broken - at least if run on OpenWrt.

What I am suggesting is indeed to remove any reliance on the definition of the syslog level implementation.
RESERVED1 and RESERVED2 are totally my invention ;-)

I said hex as the selected level would be a bitwise mask allowing us to select any combination of our own defined levels. Two hex digits will give us 8 possible levels plus "silent" and we can choose any combination of the 8 levels to be logged.

For example, if we wanted just DEBUG and WARNING then it setting level 0x24, Decimal 36, binary 00100100 will do the business given my definition above.

mwarning:
> I said hex as the selected level would be a bitwise mask allowing us to select any combination of our own defined levels

Ok, that is something i can understand. But bit masks should not be exposed to the user as debug level. Imho, It would be overkill.

bluewavenet:
Previously we had level 5 as default, level 0 as "not quite silent" and level 7 as verbose debug with any other numbers not being well defined. There was little or no documentation.

The way I am proposing is very flexible and easy to understand. Any admin user who does not understand a bit mask should not really be setting their own combined levels. We can instead document our new default of 0x0F (or Decimal 15 if you will) as the direct equivalent of the old level 5, and 0x2F (or Decimal 47) as the direct equivalent of the old level 7.

We can document the rest of it for "Advanced Users/Developers".

We already expose our admins to bit masks with fw_mark, and it has not been a problem and gives a great advantage in flexibility.

mwarning:
This is too complicated, user should not need to understand bit fields for something as debugging.
Anyway, the init script should be working again (the procd init script does not use syslog, since the client process is expected to run in foreground).

PR #393, opened 2019-07-24
put fas code into block

mwarning:
Put some fas code behind `if (config->fas_port)`.

### DIFF
--- a/src/fw_iptables.c
+++ b/src/fw_iptables.c
@@ -369,8 +369,8 @@ iptables_fw_init(void)
 	char *gw_address = NULL;
 	char *gw_iprange = NULL;
 	int gw_port = 0;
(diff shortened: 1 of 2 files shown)


mwarning:
@bluewavenet

bluewavenet:
@mwarning 
It compiles ok but fails with the log message:

> Thu Jul 25 10:44:14 2019 daemon.crit nodogsplash[15017]: safe_strdup called with NULL which would have crashed strdup. Bailing out.

mwarning:
@bluewavenet try again

bluewavenet:
That's better, seems to work ok now ;-)

mwarning:
ok, this makes the code a bit better to understand and prevents side effects from FAS code.

PR #399, opened 2019-07-29
Write log to SYSLOG. Make debuglevel 0 silent except for LOG_ERR

bluewavenet:
Write log to SYSLOG. Make debuglevel 0 silent except for LOG_ERR


bluewavenet:
@mwarning 
We did need -s in init.d file after all. Also a couple of changes to make debuglevel 0 silent as before the config is read, debuglevel defaults to 1 so any debug output before reading the config will be output according to level 1. Errors will still be logged even at level 0. I think this is what we want.
Tested on OpenWrt and all good now. Thanks for your efforts! I'll update the documentation next :-)

PR #414, opened 2019-08-13
V4.0.2beta numerous fixes and enhancements

bluewavenet:
ndsctl: json - make parsing consistent for all vars
PreAuth: Override FAS settings, update DOCS, configs and css
Fix: fas_remotefqdn not supported with option fas_secure_enabled '0'


src/util.c
```diff
@@ -690,7 +690,7 @@ ndsctl_json_all(FILE *fp)
 
 	LOCK_CLIENT_LIST();
 
-	fprintf(fp, "{\n\"client_length\": %d,\n", get_client_list_length());
+	fprintf(fp, "{\n\"client_list_length\": \"%d\",\n", get_client_list_length());
```
mwarning: why do all numbers have to be strings? This seems to be artificial.

src/main.c
```diff
@@ -272,7 +273,23 @@ main_loop(void)
...
+		debug(LOG_NOTICE, "Preauth is Enabled - Overiding FAS configuration.\n");
+		debug(LOG_NOTICE, "Preauth Script is %s\n", config->preauth);
+
+		//override all other FAS settings
+		config->fas_remoteip = safe_strdup(config->gw_ip);
+		config->fas_remotefqdn = NULL;
+		config->fas_key = NULL;
+		config->fas_port = config->gw_port;
+		safe_asprintf(&preauth_dir, "/%s/", config->preauthdir);
+		config->fas_path = safe_strdup(preauth_dir);
```
mwarning: This would be shorter and needs to temporary.
```
safe_asprintf(&config->fas_path, "/%s/", config->preauthdir);
```

src/main.c
```diff
@@ -272,7 +273,23 @@ main_loop(void)
 	/* TODO: set listening socket */
 	debug(LOG_NOTICE, "Created web server on %s", config->gw_address);
 
+	if (config->preauth) {
+		debug(LOG_NOTICE, "Preauth is Enabled - Overiding FAS configuration.\n");
```
mwarning: Why do we need to set FAS variables when it is disabled? This feels wrong.

bluewavenet: PreAuth is a FAS that uses MHD to serve html, so FAS is enabled with very specific settings. So instead of requiring the user to set them we do it for them here making it simple.

src/main.c
bluewavenet: Indeed.

src/util.c
bluewavenet: It makes parsing all the client variables sent by ndsctl very simple. The login.sh (aka demo-preauth.sh) has a very simple parser that extracts all the variables from the json output.

forward_authentication_service/PreAuth/demo-preauth.sh
```diff
@@ -150,26 +165,49 @@ fi
...
-	clientmac="$(ndsctl json $clientip | grep mac | cut -c 8- | cut -c -17)"
+	# If we got here, we have both the username and emailaddr fields as completed on the login page on the client,
+	# so we will now call ndsctl to get client data we need to authenticate and add to our log.
+
+	# Variables returned from ndsctl are listed in $varlist.
+
+	# We at least need the client token to authenticate.
+	# In this example we will also log the client mac address.
+
+	varlist="id ip mac added active duration token state downloaded avg_down_speed uploaded avg_up_speed"
```
bluewavenet: @mwarning 
json parser

bluewavenet: Sorry - That didn't work - How do you get github to display a specific bit of code?

bluewavenet: This is the json parser with check for ndsctl empty response in case of another instance of nds already running.

```
	varlist="id ip mac added active duration token state downloaded avg_down_speed uploaded avg_up_speed"
	clientinfo=$(ndsctl json $clientip)

	if [ -z $clientinfo ]; then
		echo "<big-red>Sorry!</big-red><italic-black> The portal is busy, please try again.</italic-black><hr>"
		echo -e $login_form
		echo -e $footer
		exit 0
	else
		for var in $varlist; do
			eval $var=$(echo "$clientinfo" | grep $var | awk -F'"' '{print $4}')
		done
	fi

```

src/main.c
bluewavenet: PreAuth is now enabled with the single line in config:
`	option preauth '/usr/lib/nodogsplash/login.sh'`

This is now safe to enable regardless of any previously set FAS config options. Before this change it was quite complicated to understand and very easy for the user to mess up.

PR #432, opened 2019-09-13
Bump to version 4.2.1beta

bluewavenet:


mwarning:
There is no point to create merge requests for code that is not meant to be reviewed and also consists o a single commit. It clutters the commit history. Maybe we should to a rebase...

bluewavenet:
I do not disagree with the sentiment and uncluttered is good.
But on the other hand having a clear and very simple to follow audit trail is good too.
I will try to maintain a better balance :-D

PR #440, opened 2019-09-27
Reduce memory, fix log file location, docs update

bluewavenet:
Reduce memory requirements and fix log file location.
Update documentation.

### DIFF
--- a/src/http_microhttpd.c
+++ b/src/http_microhttpd.c
@@ -609,8 +609,8 @@ static int show_preauthpage(struct MHD_Connection *connection, const char *query
 	const char *user_agent = NULL;
 	char enc_user_agent[256] = {0};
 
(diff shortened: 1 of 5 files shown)


src/http_microhttpd.c
```diff
@@ -609,8 +609,8 @@ static int show_preauthpage(struct MHD_Connection *connection, const char *query
 	const char *user_agent = NULL;
 	char enc_user_agent[256] = {0};
 
-	// Encoded querystring could be up to 3 times the size of unencoded version
-	char enc_query[QUERYMAXLEN * 3] = {0};
+	// Encoded querystring couild be bigger than the unencoded version
```
mwarning: couild => could

bluewavenet: Eeek! Fixed :-D

Issue #447, opened 2019-10-08
NDS v4.3.0 fails to create deb package on RPi 4B


martignoni:
## Steps to reproduce

```bash
git clone https://github.com/nodogsplash/nodogsplash.git
cd nodogsplash/
git checkout v4.3.0
make deb
```

## Error occurring

```
cp: cannot create regular file '/home/xxx/nodogsplash/dist/nodogsplash-4.3.0/debian/nodogsplash/usr/lib/nodogsplash/login.sh': No such file or directory
make[2]: *** [Makefile:45: install] Error 1
```

## Note

On the same hardware, NDS v4.2.0 creates deb package flawlessly (just typing `git checkout v4.2.0` instead of `git checkout v4.3.0`.

bluewavenet:
@martignoni
Thanks for reporting this. 
I see the problem - a missing line in the Debian makefile, it does not make the library directory.
Fix on its way...

bluewavenet:
@martignoni 
see commit https://github.com/nodogsplash/nodogsplash/commit/3c4013133353f43ff8bb14c4ea33e0a63b412ff0

Could you please test by add the missing mkdir line to your Makefile?
If this works I'll tag a new release.

martignoni:
Yes, the deb package creation works with the fix indicated. 👍

bluewavenet:
@martignoni 
Thanks ;-)
https://github.com/nodogsplash/nodogsplash/releases/tag/v4.3.1

martignoni:
Thank **you**!

martignoni:
Ooops:
```
dpkg-source: error: can't build with source format '3.0 (quilt)': non-native package version does not contain a revision
```

bluewavenet:
@martignoni 
The source for libmicrohttpd has been intermittently offline today. I think that is the problem.
The Travis checks were failing for the same reason. Keep trying, will probably return to normal very soon.

bluewavenet:
The primary mirror https://mirrors.ocf.berkeley.edu/gnu/ is offline or extremely slow at the moment (Wed Oct 9 18:35 UTC). 
There will be some poor server engineer wishing he had taken the day off I suspect ;-)

martignoni:
Indeed. Thanks for the info, I'll try tomorrow.

martignoni:
Cause was a wrong version number. Just created a PR to fix it.

martignoni:
(Sorry, I should have raised another issue for this.)

[...]

PR #450, opened 2019-10-10
Fix deb package build error

martignoni:
- See https://github.com/nodogsplash/nodogsplash/issues/447#issuecomment-540098767.


bluewavenet:
@martignoni 
Well spotted thankyou!

Issue #515, opened 2020-02-03
libmicrohttpd version


bluewavenet:
@lynxis @mwarning 
The current libmicrohttpd package version in OpenWrt 19.07.1 and Debian is v0.9.62

Query string data sent to NDS by the client browser is always url escaped.

libmicrohttpd (MHD) should then send elements of the received escaped query string for unescaping (either to its builtin unescape or unescape call back)

However MHD v0.9.62 selectively unescapes  certain special characters before calling the unescape function resulting in lost or potentially corrupted characters. ie:

 * Character "$": browser escapes to %24, MHD unescapes %24 to $ **_BEFORE_** calling its configured unescape function, resulting in a double unescape

 * Character "%": browser escapes to %25, MHD unescapes %25 to % **_BEFORE_** calling its configured unescape function, resulting in a double unescape

 * Character "&": browser escapes to %26, MHD unescapes %26 to a space **_BEFORE_** calling its configured unescape function, resulting in loss of information

 * Character "+": browser escapes to %2B, MHD unescapes %2B to a space **_BEFORE_** calling its configured unescape function, resulting in loss of information

This effects form data submitted by BinAuth and PreAuth as MHD is used as the web server.
It does not effect FAS.
**!!EDIT!!** It could indirectly effect FAS if BinAuth is used for local post authentication processing.

At the time of writing, the current release of libmicrohttpd is v0.9.69.
This is also the current package version in OpenWrt Snapshot.

**_libmicrohttpd v0.9.69 does not exhibit the above incorrect behaviour._**

MHD has the constant MHD_VERSION, but this is only the value contained in the microhttpd.h that is included when the NDS package is compiled.

It does **_NOT_** return the version of libmicrohttpd that is **_ACTUALLY_** installed.

The minimum required version of MHD is set in main.c, but this does not prevent a compiled NDS package being used with an older version of MHD. It will however work in the case of a `make install` in Debian/done on the actual target device eg Rpi etc. 

**_Is there any way you can think of for NDS to detect the ACTUAL installed version of MHD?_**

PR #521, opened 2020-02-09
Force minumum version of libmicrohttpd (MHD)

bluewavenet:
See issue #515

The setting in main.c works only if NDS is compiled and installed in place.

If creating a package however, microhttpd.h is embedded as an include
leaving no way for NDS to discover the installed MHD version.

This commit uses the NDS init.d script on Openwrt to determine the
installed MHD version.

opkg cannot be used to determine the MHD version as on installing (with opkg)
an opkg lock error is generated.

If in the future, NDS was to be added to other Linux official package repo's
(eg Debian), similar installed version detection would have to be done.

### DIFF
--- a/openwrt/nodogsplash/files/etc/init.d/nodogsplash
+++ b/openwrt/nodogsplash/files/etc/init.d/nodogsplash
@@ -113,6 +113,43 @@ generate_uci_config() {
   local download
   local upload
 
+# Check libmicrohttpd version
+  local minversion="0.9.69"
+  local oldver="libmicrohttpd is out of date, please upgrade to at least version $minversion, Exiting...."
+
+  if [ -f "/usr/lib/opkg/info/libmicrohttpd-no-ssl.control" ]; then
+    local cpath="/usr/lib/opkg/info/libmicrohttpd-no-ssl.control"
+
+  elif [ -f "/usr/lib/opkg/info/libmicrohttpd-ssl.control" ]; then
+    local cpath="/usr/lib/opkg/info/libmicrohttpd-ssl.control"
+
+  else
+    echo "libmicrohttpd is not installed, Exiting...."
+    return 1
+  fi
+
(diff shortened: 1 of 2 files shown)


bluewavenet:
@lynxis @mwarning 
see issue #515 

Travis is failing. Is this because the Travis system does not have the minimum required MHD version?

openwrt/nodogsplash/files/etc/init.d/nodogsplash
```diff
@@ -113,6 +113,43 @@ generate_uci_config() {
...
+
+  elif [ -f "/usr/lib/opkg/info/libmicrohttpd-ssl.control" ]; then
+    local cpath="/usr/lib/opkg/info/libmicrohttpd-ssl.control"
+
+  else
+    echo "libmicrohttpd is not installed, Exiting...."
+    return 1
+  fi
+
+  local min_major=$(echo "$minversion" | awk -F '.' '{print $1}')
```
mwarning: openwrt has newer_than as shell command. That might simplify some things here.

mwarning:
libmicrohttpd has MHD_get_version to get the runtime version of libmicrohttpd.

bluewavenet:
@mwarning 
I can't find any reference to MHD_get_version. Any pointers?

mwarning:
This works fine:
```
#include <stdlib.h>
#include <stdio.h>
#include <microhttpd.h>

int main(int argc, char *argv[])
{
  int major = 0;
  int minor = 0;
  int patch = 0;
  const char *version = MHD_get_version();
  if (sscanf(version, "%d.%d.%d", &major, &minor, &patch) == 3) {
    printf("%d %d %d\n", major, minor, patch);
  }
  return 0;
}
```

bluewavenet:
@mwarning 
Ah! Cunningly disguised as a function :-D
I'll give it a try.

bluewavenet:
Closing. Using MHD_get_version() instead.

PR #522, opened 2020-02-10
Add option use_outdated_mhd

bluewavenet:
See issue #515

MHD_get_version is used to determine the installed (runtime) version of
libmicrohttpd (MHD).
If the version is earlier than 0.9.69, then by default, NDS will terminate.

However, if option use_outdated_mhd is set to 1, NDS starts normally
but logs an error.

### DIFF
--- a/openwrt/nodogsplash/files/etc/config/nodogsplash
+++ b/openwrt/nodogsplash/files/etc/config/nodogsplash
@@ -33,6 +33,15 @@ config nodogsplash
 	#
 	option login_option_enabled '0'
 
+	# Use outdated libmicrohttpd (MHD)
+	# Older versions of MHD convert & and + characters to spaces when present in form data
+	# This can make a PreAuth or BinAuth impossible to use for a client if form data contains either of these characters
+	# eg. in username or password
+	# MHD versions earlier than 0.9.69 are detected.
+	# If this option is set to 0 (default), NDS will terminate if MHD is earlier than 0.9.69
+	# If this option is set to 1, NDS will start but log an error.
+	option use_outdated_mhd '0'
+
 	# MHD Unescape callback
 	# MHD has a built in unescape function that urldecodes incoming queries from browsers
 	# This option allows an external unescape script to be enabled
--- a/src/main.c
+++ b/src/main.c
@@ -247,6 +247,7 @@ main_loop(void)
 	int major = 0;
 	int minor = 0;
 	int patch = 0;
+	int outdated = 0;
 	const char *version = MHD_get_version();
 
 	debug(LOG_NOTICE, "MHD version is %s", version);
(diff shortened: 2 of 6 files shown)


openwrt/nodogsplash/files/etc/config/nodogsplash
```diff
@@ -33,6 +33,15 @@ config nodogsplash
 	#
 	option login_option_enabled '0'
 
+	# Use outdated libmicrohttpd (MHD)
+	# Older versions of MHD convert & and + characters to spaces when present in form data
+	# This can make a PreAuth or BinAuth impossible to use for a client if form data contains either of these characters
+	# eg. in username or password
+	# MHD versions earlier than 0.9.69 are detected.
+	# If this option is set to 0 (default), NDS will terminate if MHD is earlier than 0.9.69
+	# If this option is set to 1, NDS will start but log an error.
```
mwarning: A warning instead of an error sounds better.

src/main.c
```diff
@@ -235,6 +243,47 @@ main_loop(void)
...
+	int major = 0;
+	int minor = 0;
+	int patch = 0;
+	const char *version = MHD_get_version();
+
+	debug(LOG_NOTICE, "MHD version is %s", version);
+
+	if (sscanf(version, "%d.%d.%d", &major, &minor, &patch) == 3) {
+
+		if (major < MIN_MHD_MAJOR) {
```
mwarning: Please shorten the code and put it into it's own method.

openwrt/nodogsplash/files/etc/config/nodogsplash
bluewavenet: Yes it does, but I wanted it to show in the log even if debuglevel is set to 0 (silent).
Updating on OpenWrt preserves the config, so if debuglevel _**is**_ set to 0 it will not be seen.

src/main.c
bluewavenet: Mmm, yes, a bit repetitive.

mwarning: Yes, that is better. But please but the logic in a separate method. Either a method to get the MHD version or a method to check the version:
```
// return libmicrohttpd version in hex format (same as MHD_VERSION)
unsigned get_microhttpdversion() {
  unsigned major = 0;
  unsigned minor = 0;
  unsigned patch = 0;
  const char *version = MHD_get_version();
  if (sscanf(version, "%x.%x.%x", &major, &minor, &patch) == 3) {
    return (major << 24) | (minor << 16) | (patch << 8);
  } else {
     return 0;
  }
}

int
main(int argc, char *argv[])
{
  unsigned required_version = 0x00096900;
  unsigned found_version = get_microhttpdversion();
  if (found_version < required_version) {
    printf("too old\n");
  }
  return 0;
}
```

openwrt/nodogsplash/files/etc/config/nodogsplash
mwarning: ok

src/main.c
bluewavenet: Well, it is either:
 * split up into major, minor patch and do up to 3 tests
 * split up into major, minor patch, convert to hex format then do 1 test

Also, we only need to check once on startup so there is little point in using a function.

Yours is smaller than mine (but mine logs the required version) ;-)

As I am very short of time I would like to leave it as it is.

Note: I kept the original check for MHD_VERSION at 0x00095100 as 0.9.51 is the minimum required microhttpd.h for NDS to be compiled with. As we know, this is not related in any way to the MHD version that may be available at runtime.
0.9.51 is good enough if using only an external FAS, but can be problematic with form data if using MHD to serve the client page(s).

mwarning: My point was about to put the code to acquire the version into a separate method. This way it does not pollute the namespace of the main() method.

bluewavenet: Having looked up "namespace pollution", I see what you mean.
main_loop() as it now stands is badly named, it should be more like "setup_then_start_main_loop".

I think we should do some more general housekeeping in the near future, as the subject of one or more new PRs

PR #532, opened 2020-02-26
Enable https protocol for remote FAS

bluewavenet:
Previously a remote FAS could not use https protocol without generating
browser security errors or warnings so NDS enforced use of http.

A new FAS level, fas_secure_enabled = 3 is introduced here.

Level 3 is the same as level 2 except the use of https protocol is
enforced for FAS. In addition, the "authmon" daemon is loaded.

This daemon allows the external FAS, after client verification is complete,
to effectively traverse inbound firewalls and address translation to
achieve NDS authentication without generating browser security warnings.
or errors.

A fully functional https fas script is provided (fas-aes-https.php).

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -136,6 +136,7 @@ void
 termination_handler(int s)
 {
 	static pthread_mutex_t sigterm_mutex = PTHREAD_MUTEX_INITIALIZER;
+	char *fasssl = NULL;
 
 	debug(LOG_NOTICE, "Handler for termination caught signal %d", s);
 
@@ -146,6 +147,12 @@ termination_handler(int s)
 	} else {
 		debug(LOG_INFO, "Cleaning up and exiting");
 	}
+
+	// Check if authmon is already running and if it is, kill it
+	debug(LOG_INFO, "Explicitly killing the authmon daemon");
+	safe_asprintf(&fasssl, "kill $(pgrep -f authmon) > /dev/null 2>&1");
+	system(fasssl);
+	free(fasssl);
 
 	auth_client_deauth_all();
 
@@ -230,9 +237,12 @@ main_loop(void)
 	int result = 0;
 	pthread_t tid;
 	s_config *config;
+	char protocol[8] ={0};
 	char msg[255] = {0};
+	char gwhash[255] = {0};
 	char *fasurl = NULL;
 	char *fasssl = NULL;
+	char *gatewayhash = NULL;
 	char *fashid = NULL;
 	char *phpcmd = NULL;
 	char *preauth_dir = NULL;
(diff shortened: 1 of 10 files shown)


bluewavenet:
@mwarning 
After this is merged, prior to a new release, I intend to do a tidy-up of main.c to eliminate the namespace pollution you pointed out previously.

src/main.c
```diff
@@ -442,6 +453,68 @@ main_loop(void)
...
+			}
+			free(fasssl);
+
+			// Start authmon in the background
+			safe_asprintf(&fasssl,
+				"/usr/lib/nodogsplash/authmon.sh \"%s\" \"%s\" \"%s\" &",
+				config->fas_url,
+				gatewayhash,
+				config->fas_ssl
+				);
```
mwarning: Please indent the closing parenthesis on the column of the function name. I think that is what do everywhere else.

src/main.c
```diff
@@ -442,6 +453,68 @@ main_loop(void)
...
+				debug(LOG_ERR, "authmon daemon failed to load\n");
+				debug(LOG_ERR, "Exiting...");
+				exit(1);
+			} else {
+				debug(LOG_NOTICE, "authmon daemon is loaded\n");
+			}
+
+			free(fasssl);
+			free(gatewayhash);
+}
```
mwarning: needs indentation

src/ndsctl_thread.c
```diff
@@ -313,6 +318,17 @@ ndsctl_auth(FILE *fp, char *arg)
 	}
 	UNLOCK_CLIENT_LIST();
 
+	/* set client values */
+	client->download_limit = download;
```
mwarning: Why not assign zero here instead of using a dummy variable?
Same for upload and now.

src/ndsctl_thread.c
```diff
@@ -295,9 +295,14 @@ ndsctl_handler(int fd)
 static void
 ndsctl_auth(FILE *fp, char *arg)
 {
+	s_config *config = config_get_config();
 	t_client *client;
 	unsigned id;
 	int rc;
+	int seconds = 60 * config->session_timeout;
```
mwarning: Is this meant to mean forever?
I think -1 is a value for that.

src/main.c
bluewavenet: On checking, it seems it is not done everywhere else, but I agree it should be.

src/main.c
bluewavenet: Oops

src/ndsctl_thread.c
bluewavenet: The intention is to integrate binauth functionally as a next enhancement. The values of these variables will be set by the binauth script.
I will add a TODO comment here for now.

src/ndsctl_thread.c
bluewavenet: 0 would be forever, but we want this to be the value of client->session_end in seconds.

bluewavenet:
Mmm. I didn't close it, I merged it.... Very odd. I think I had a "Cat walking on keyboard" error :-D

PR #535, opened 2020-03-01
Tidy up main_loop, removing namespace pollution

bluewavenet:

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -229,15 +229,13 @@ init_signals(void)
 }
 
 /**@internal
- * Main execution loop
+ * Setup from Configuration values
  */
 static void
-main_loop(void)
+setup_from_config(void)
 {
-	int result = 0;
-	pthread_t tid;
-	s_config *config;
-	char protocol[8] ={0};
+	char protocol[8] = {0};
+	char port[8] = {0};
 	char msg[255] = {0};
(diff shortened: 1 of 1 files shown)


src/main.c
```diff
@@ -355,10 +359,10 @@ main_loop(void)
 			exit(1);
 		}
 	}
-
 	/* TODO: set listening socket */
 	debug(LOG_NOTICE, "Created web server on %s", config->gw_address);
 
+	/* If login script is enabled, check if the script actually exists */
```
mwarning: We should do one line comments in one style. `/**/` or `//`

src/main.c
```diff
@@ -229,15 +229,13 @@ init_signals(void)
 }
 
 /**@internal
- * Main execution loop
+ * Setup from Configuration values
  */
 static void
-main_loop(void)
+setup_from_config(void)
```
mwarning: Maybe we should pass the config variable here as an argument.
I do not have a strong opinion here.

src/main.c
bluewavenet: Yes, I agree. As I am in favour of having lots of comments, I think having a recommended style is a good thing. My preferred use is // for single line comments.

mwarning: then let's use `//` from here on :)

PR #538, opened 2020-03-03
Coding style update - use // for single line comments

bluewavenet:

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -39,13 +39,13 @@
 #include <sys/stat.h>
 #include <arpa/inet.h>
 
-/* for strerror() */
+// for strerror()
 #include <string.h>
 
-/* for wait() */
+// for wait()
 #include <sys/wait.h>
 
-/* for unix socket communication*/
+// for unix socket communication
 #include <sys/socket.h>
 #include <sys/un.h>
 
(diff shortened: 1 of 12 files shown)


src/main.c
```diff
@@ -191,12 +193,14 @@ init_signals(void)
...
-	/* Trap SIGPIPE */
+	// Trap SIGPIPE
+
 	/* This is done so that when libhttpd does a socket operation on
 	 * a disconnected socket (i.e.: Broken Pipes) we catch the signal
 	 * and do nothing. The alternative is to exit. SIGPIPE are harmless
 	 * if not desirable.
 	 */
+
 	debug(LOG_DEBUG, "Setting SIGPIPE  handler to SIG_IGN");
```
mwarning: an extra line sneaked in here :)

src/template.c
```diff
@@ -47,20 +47,21 @@ int tmpl_parse(struct template *vars, char *dst, size_t dst_len, const char *src
 			continue;
 		}
 
-		/* we know it's a '$'. But we are interest in the next char */
```
mwarning: Let's fix typos in a separate commit.

src/main.c
bluewavenet: It was deliberate ;-)
As it is an inline comment, the extra line adds to visual balance and readability in my opinion.

src/template.c
bluewavenet: Maybe, but I would prefer to fix little typos in comments as I spot them, otherwise I am sure to forget them later.

PR #545, opened 2020-03-08
Update documentation about libmicrohttpd

martignoni:


bluewavenet:
@martignoni 
See:
Version 5.0.0 - FAS API Moved to OpenNDS #560 
I would much appreciate it if you could check:
https://opennds.readthedocs.io
and resubmit on openNDS if required.

lynxis:
I've to check if we're okay again with older libmicrohttpd version. With the drop of the FAS we should have lowered the requirements of  libmicrohttpd.
I'm closing this PR for now, since it was against nodogsplash 4.x. and the master moved on to 5.x.

bluewavenet:
@lynxis 
The older versions of MHD have problems with the built in unescape where some special characters are not unescaped correctly.
This effected any strings passed in and out such as usernames, passwords and the gatewayname.
Without FAS and BinAuth, this leaves just gatewayname, so will be more of an obscure edge case.

PR #558, opened 2020-04-10
Strip down to templated splash page only.

bluewavenet:
Create a version optimised for minimal resource utilisation, supporting only the simple templated splash page.

The forwarding authentication service components have been moved to the new openNDS package at:
https://github.com/openNDS/openNDS


bluewavenet:
@mwarning 
@lynxis

lynxis:
Thanks. I've already started removing the FAS. I've to take a closer look into your documentation commit. But we can close this PR for now, because might cherry-pick smaller parts of it.

Issue #328, opened 2018-11-19
Running NDS without Internet Access


bluewavenet:
Running NDS without Internet Access
or more precisely 
NDS Automatic Offline Support for Client CPD

**The Issue:**
Client CPD does not trigger the NDS captive portal if the CPD fails to resolve the ip address of its connectivity check URL.

NDS can be made to function in an offline situation by setting up the router DNS server to return an ip address, eg the gateway address, for ALL DNS requests rather than returning an error.

This is fine for a situation where the router never has an Internet feed, but of no use if the feed is only temporarily unavailable.

**Proposed Solution:**
NDS redirects all port 53 traffic from the local network to its own DNS handler at a different port, say 5353.

An NDS DNS handler then calls the  router's actual DNS server on port 53. If it receives an ip address, it echoes this to the client. If it receives an error, ie the router is offline, then it sends its own gateway address to the client.

The client CPD will then access the ip address it receives on port 80 thus triggering the captive portal.

As NDS now knows that it is operating offline it can send an "Offline Status" splash page instead of the usual splash page. 

This should be fairly simple to implement.

In addition, by using FAS and staying on the splash page without actually authenticating to NDS, this automated behavior can be used to provide offline information services, getting round the issue with the defacto standard CPD automatically closing the client CPD browser on authentication.

@mwarning

mwarning:
Won't it be sufficient to return the gateways IP address if the router is meant to work offline?
This is already possible on OpenWrt by setting: `uci add_list dhcp.@dnsmasq[0].server='/p2p/::1#3535'` (we should include that into the NDS config on OpenWrt..)

mwarning:
Ok, you want to have a dynamic detection. If there is not Internet => return Gateway IP, if there is Internet => return gateway IP if unauthenticated and real IP if authenticated.
Is that correct?

bluewavenet:
Not quite. 
1. If there is Internet => client CPD DNS query for connectivity check url gets real IP - everything as normal - splash page is triggered.
2. Internet Down (or not even available) => client CPD DNS query for connectivity check url gets gateway address - splash page is triggered.

In case 2, if the CPD gets NACK or error from DNS server, it stops trying to connect and informs Networkmanager to try another SSID. Splash page is never shown.

bluewavenet:
Once authenticated, the same applies. 
Internet present - real ip address returned from DNS query
Internet not present - gateway address returned

bluewavenet:
[...]

PR #592, opened 2020-10-15
Fake DNS for pre-auth clients

thormj:
Added a fakedns thread that simply returns the ip4 address to *any* dns request, and if that is enabled, redirects all dns traffic from pre-auth users to it.  If you use this, the gateway must know its hotspot IPV4 address (the DNS server only handles IPV4).

Makes it much faster to display the splash page (the dns queries don't go to the internet), makes offline use easy, and makes the system iodine-proof.


bluewavenet:
@thormj 
I have not looked into your code in any detail, but given it is the client CPD that triggers the splash page, there are some fundamental issues to take into account.

- An increasing number of vendor CPD implementations implement their own DNS Rebind protection by discarding upstream RFC1918 responses. The result is that such CPD implementations will not trigger the splash page in this case. You can work around it by returning a valid public ip address for all requests, rather than a private one. The returned address can be a random one, for example 123.123.123.123.
- Most CPD implementations and browsers cache DNS responses with a TTL override, some caching indefinitely, or at least until a restart. In this case, a "fake DNS" will result in cache poisoning. This does not matter much for continuous offline operation, but will be bad if the fake is enabled for online operation.

mwarning:
@thormj thanks, awesome contribution. I will try to give it a test run. Possible next week.

thormj:
Man... they're really trying to screw up a good thing (automatically finding the "page that enables the network"... trust me, the code is naive (I don't even bother to look at the DNS query).

For chrome, there's an internals page that resets things - https://stackoverflow.com/questions/4322046/prevent-chrome-from-caching-a-sites-dns-records/39124052#39124052

I've updated it to have a few more options that should make it as flexible as we'd want to make it:
predns_port [5553] - Set to 0 to not do the iptable redirection.
predns_ip [NULL] - if null, it uses the gw_ip4 address...
predns_internal [1] - 0 to not run the naive internal dns server
predns_internalresp [NULL] - the ipv4 response of the internal server; if null, returns gw_ip4 address.

Now it should be flexible enough to do just about anything; if you want to set up a "preath dnsmasq server", set predn_internal=0, and predns_ip : predns_port to the "preauth dnsmasq server" (then you can configure the resolv tables however you want.

If you set predns_port = 0 no firewall rules will be written for you (and the server won't be started because it's kinda useless).
If you leave everything as a default, it sets up its own naive dns server on 5553 pointing to the gw,

thormj:
> @thormj thanks, awesome contribution. I will try to give it a test run. Possible next week.

So I don't really know etiquette... I've made it more flexible in response to @bluewavenet (so now it can point to whatever you want as a pre-auth-dns)...

Wasn't sure if I should let you run with this (evidently a pull request keeps pointing to the latest head on the branch),
 -or-
Close and make a new request (don't use the ones before Removed a Couple of Warnings -- I was guilty of using git as a "file mover" for this.

mwarning:
If we merge this, we would probably squash all commits into one. You can do it yourself as well (`git rebase - i ...`), if you like.

bluewavenet:
@thormj @mwarning 
Yes excellent work. But I have a number of concerns.

> Man... they're really trying to screw up a good thing

**Not really.** It is all about security and performance. All browsers and all client operating systems of any note implement a built in CPD to mitigate conflicts between client security and the necessity to support captive portal situations. These CPD "browsers" clean up and possibly self terminate once the detection URL returns the expected response.

For some insight into CPD, see:
https://opennds.readthedocs.io/en/stable/howitworks.html#captive-portal-detection-cpd

However, implementing a captive portal with a fake DNS turns it into a "DNS Rebind attack tool", even if we do not intend it to be used as such. New releases of CPD implementations are enabling detection of DNS rebind and rejecting if detected.

**The original thinking** behind this was to provide an "Offline Mode" with the further possibility of auto detecting the failure of the Internet feed ( #328 ).

**_The important bit of functionality missing in this PR is described in #328 :_**

> [The] NDS DNS handler then calls the router's actual DNS server on port 53. If it receives an ip address, it echoes this to the client. If it receives an error, ie the router is offline, then it sends [some valid public ip] to the client.

Note the update to the quote in the light of recent changes in CPD implementation.

**This PR would also need** a rationalisation of the new config parameters including a clear naming scheme (in my opinion), but we can go into that later.

**You must also not forget** to update the OpenWrt config files as well as the documentation and man pages ;-)

**My final concern** is that this adds complexity to NoDogSplash, shortly after a deliberate forking to minimise complexity.....

thormj:
Updated the OpenWRT docs... I didn't really see anything in the man page talking about the config (more about how to integrate it into a system), so let me know if I missed it.

If you want to return a public IP, you can use the PreDNS_InternalResp to whatever IPv4 you'd like (8.8.8.8 ?).
The caching has bitten me with this, but I can't figure out a good way to fix it (if they'd simply obey the TTL!); I suppose you could make a list of "detection addresses" and then configure dnsmasq to do something "special" with them and return NXDomain for everything else.

I do understand why they're doing it (there's an evil-network-on-a-stick that's amazing -- otoh, it /doesn't/ pop up a CPD page, it just supersedes your DNS, etc with its own evil interfaces and uses your browser's tendencies to reconnect everything [server-side-pushes, etc] to harvest passwords and such)...

I put it some notes in the fakedns_thread.c file -- I thought I'd like to be able to pass in -1.0.0.0 and have it simply return nxdomain for everything of -2.0.0.0 have it return nodata, but I think it'd have to be split out so it understood the names because we need to return /something/ for the CPD check names... but that's getting even more complicated (and maybe a job for a "properly configured dnsmasq")...

I'll think about making the internal server more like #328... it seems like a good way to handle semi-online mode (in my case, this RasPi node is my CNC controller out in the shed; sometimes it can connect to the main network, but if all the machines are on, it doesn't do well; using HostAPD means my phone can *reliably* control the CNC and /sometimes/ see the internet.  The NDS was added so when it connects it automatically pops up the CNC page without having to have a shortcut (this is a MaslowCNC/Webcontrol thing)...

I have a number of "offline wifi devices" (eg Twinkly), but since they don't have any CPD splash page, you really don't know what to do if you "connect" to them; this is my attempt to give WIFi IOT devices an easy "splash page" (vs the "I am 8.8.8.8" hack on https://stackoverflow.com/questions/54583818/esp-auto-login-accept-message-by-os-with-redirect-to-page-like-public-wifi-port

bluewavenet:
@thormj 

> Updated the OpenWRT docs

**It is not OpenWrt docs but essential procd and config files for OpenWrt**.  Note that OpenWrt is the major platform for NDS in both its forms (NoDogSplash and openNDS).

**The documentation** is a set of .rst files that are exported to readthedocs.io
ie https://nodogsplashdocs.readthedocs.io/en/stable/

**The man pages** are a duplicate of this. Both need to be updated to reflect changes.

> If you want to return a public IP, you can use the PreDNS_InternalResp to whatever IPv4 you'd like (8.8.8.8 ?).
> The caching has bitten me with this

**You are missing the point** that fake dns should only be in operation **IF** and only if, you are running offline.

**Automatic detection of an offline state** when a client connects (ie the Internet feed went down), would be quite easy to implement with your fake server, as I described earlier. This would allow clients to be informed that there is no Internet access.

> I have a number of "offline wifi devices"

**All you need for permanently offline devices** is the trivial one line edit to dnsmasq config.

**Your fake server has the potential to be very useful**, but your overall implementation seems to be biased towards your limited use case, resulting in the discussed problems.

**I see its main benefits for:**
 * Providing a simple "Offline" or "Museum" mode for _permanent offline operation_ where clients are served an information web site. Enabled by a simple NDS config option. This works extremely well at museum type venues with QR codes at each exhibit, for example.
 * Providing  _automated offline detection with failover_ to a local web server giving, in its simplest form, an apology message to clients, and at the other extreme a full informational or campus wide web service.

**I do not intend to be "down voting" with my comments,** I'm just trying to put forward the larger picture - in my opinion it is just adding unnecessary complexity to NoDogSplash where a simple one liner in the dnsmasq config would suffice.

Cherry picking this over to in openNDS though, is a different matter, as the two main benefits above can really be made to work simply and effectively.

thormj:
Hrm... it looks like something _should_ be done with the fake server.  I was very happy with it when I use it from a Windows PC (only one source of internet, so it connects and goes)., but my phone [Galaxy S7E] doesn't like it at all.  It doesn't tell the user what's going on other than "we'll connect to this when it's connected"... even if you start the phone in Airplane mode, it refuses to bring up the splash page until the  fake server is turned off.

I'm playing around with dnsmasq and seeing what I can do to troubleshoot/fix the issue.... maybe a O114 on the DHCP line and some kind of "preloaded" cache for certain DNS entries...  but I only have Windows, GalaxyS7E, and a iPhone 6c to test with.

thormj:
The other spooty thing the GX7E does is just-close-the-window on authentication... which means the user has no way to know/get back to the gateway...

thormj:
Ehh... if you simply let dnsmasq supply good dns addresses  for the detection urls, it works like a champ.  Doesn't prevent iodine, but it works offline or even semi-offlne. 

Use #593  instead...

bluewavenet:
@thormj 

> the GX7E does is just-close-the-window on authentication

This is increasingly the way it is with vendor CPD implementations. The ratio of just-closes to keep-opens is currently significantly >1 for OS/NetworkManager CPD, but approx 0 for browsers. eg Firefox (mostly) keeps the tab open.
This is the reason RedirectURL is deprecated in openNDS.

bluewavenet:
@thormj 

> my phone [Galaxy S7E] doesn't like it at all. It doesn't tell the user what's going on other than "we'll connect to this when it's connected"... even if you start the phone in Airplane mode, it refuses to bring up the splash page until the fake server is turned off.

See my earlier comment:

> However, implementing a captive portal with a fake DNS turns it into a "DNS Rebind attack tool", even if we do not intend it to be used as such. New releases of CPD implementations are enabling detection of DNS rebind and rejecting if detected. 

You can work around this by using a **_NON_** RFC1918 address as the fake returned address. eg 123.123.123.123 works well.

bluewavenet:
@thormj 

> if you simply let dnsmasq supply good dns addresses for the detection urls, it works like a champ. Doesn't prevent iodine

For the majority of free public wifi installations, iodine abuse probably is a waste of time for the person trying to do it, as, well, access is free, so why bother trying. I suspect the only successful public wifi systems from now on, will be the free ones (this is already true in Europe apart from a few legacy ones that are hanging on with a rapidly diminishing number of customers). In countries with expensive Internet or expensive/poor/non-existent 4G, there is still a market for paid public wifi, but this will change, probably very rapidly.
So yes an iodine trap would be nice. A modified forwarding implementation of your fakedns_thread.c could check for iodine before forwarding DNS requests to dnsmsaq.

dtbayles:
@thormj 
> if you simply let dnsmasq supply good dns addresses for the detection urls, it works like a champ. Doesn't prevent iodine

What's the 1 line fix to add to `dnsmasq`? 

I require a "Museum" type solution that @bluewavenet described but for museum's in West Africa with unreliable network. If we don't have to rely on a live connection, then we don't want to. I was thinking about mocking a DNS server and sure enough here is your PR. The networking got a bit too technical for me at the end. What would be the simplest (and best) solution to display a Captive Portal to users who will always be in pre-auth and not require internet connection? Our landing page is a `.php` file, would this also be possible in the Captive Portal or must we convert our solution to just html/css/js?

bluewavenet:
@dtbayles 

> What's the 1 line fix to add to dnsmasq?

Make it **resolve everything** **_to a random non rfc1918 address._**

> Our landing page is a .php file,

NoDogSplash does not support php, for that you should migrate to openNDS.
But be aware that you must not authenticate at any time as the CPD browser window will close if you do.

> must we convert our solution to just html/css/js?

The client based CPD browser, for security reasons, supports only a very limited set of simple javascript. Any "fancy js scripting" will inevitably fail on all mobile devices.

PR #593, opened 2020-10-21
Offline Hints

thormj:
Instead of a FakeDNS server, these are instructions to add to your _host_ file to do offline operation.
I also added the FASpath/FASport to the Readme.md as an example of the difference between OpenNDS and Nodogsplash.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,6 +4,7 @@ Nodogsplash is a Captive Portal that offers a simple way to provide restricted a
 
 It was derived originally from the codebase of the Wifi Guard Dog project.
 
+
 Nodogsplash is released under the GNU General Public License.
 
 * Mailing List: http://ml.ninux.org/mailman/listinfo/nodogsplash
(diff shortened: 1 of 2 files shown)


bluewavenet:
@thormj 
There is much more to openNDS than just FASpath/FASport and for people not familiar with it FASpath/FASport will be meaningless. For better clarity, perhaps it should say something like:

> If you want a more sophisticated authentication system providing a dynamic web interface, you need openNDS with its Forwarding Authentication Service API (FAS), rather than NoDogSplash.

- making sure the link on the word "openNDS" is maintained.

bluewavenet:
@thormj 
Adding entries to the hosts file might seem to be a good idea at first sight, but you will find there are many, perhaps hundreds of CPD detection urls. Not only that, they have also been known to change with vendor updates as well as varying by model. Samsung is one of the worst for this. Chromebooks from various manufacturers as well.

I don't mean to discourage you, on the contrary, it is very refreshing to see someone going down the same rabbit hole I went down a few years ago :-D Someone else going through the same thought processes is very valuable.

Adding to the hosts file might be a good solution for edge cases, for example IoT management using a standardised "engineers" phone for example.

With regard to "Hints", in my opinion the correct place is in the documentation. A new section would be most useful.
By documentation, I mean in the docs folder of course.

PR #631, opened 2023-09-16
Fix buffer overflow

mwarning:
Fix a buffer overflow for long queries.

Script to trigger

```python
#!/usr/bin/env python3

import socket 

CAPTIVE_IP = '192.168.99.3'
CAPTIVE_PORT = 2050

overflow = 'A'*30000
r = f'GET /?host=kek{overflow} HTTP/1.1\nHost: localhost\n\n'
sock = None
try:
    sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
    sock.connect((CAPTIVE_IP, CAPTIVE_PORT))
    sock.send(r.encode())
finally:
    if sock != None:
        sock.close()
````

### DIFF
--- a/src/http_microhttpd.c
+++ b/src/http_microhttpd.c
@@ -600,7 +600,9 @@ static int encode_and_redirect_to_splashpage(struct MHD_Connection *connection, 
 
 	if (originurl) {
 		if (uh_urlencode(encoded, sizeof(encoded), originurl, strlen(originurl)) == -1) {
-			debug(LOG_WARNING, "could not encode url");
+			debug(LOG_WARNING, "cannot encode url");
+			/* not enough memory */
+			return send_error(connection, 503);
 		} else {
 			debug(LOG_DEBUG, "originurl: %s", originurl);
 		}
--- a/src/http_microhttpd.c
+++ b/src/http_microhttpd.c
@@ -14,8 +14,8 @@
  * @file http_microhttpd.c
  * @brief a httpd implementation using libmicrohttpd
  * @author Copyright (C) 2015 Alexander Couzens <lynxis@fe80.eu>
+ * @author Copyright (C) 2023 Moritz Warning <moritzwarning@web.de>
  */
-
 
 #include <microhttpd.h>
 #include <syslog.h>
@@ -27,6 +27,7 @@
 #include <linux/limits.h>
 #include <fcntl.h>
 #include <unistd.h>
+#include <stdbool.h>
 
 #include "client_list.h"
 #include "conf.h"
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by mwarning** [comment]
@lynxis ping

**Comment by mwarning** [comment]
@ standash fyi

**Comment by mwarning** [comment]
@lynxis is it OK for you to use stdbool.h?

File: src/http_microhttpd.c [hunk 1]
```diff
@@ -629,25 +635,29 @@ static int encode_and_redirect_to_splashpage(struct MHD_Connection *connection,
...
+		.capacity = QUERYMAXLEN,
+		.query = query,
+	};
+
+	// collect query
+	if (MHD_get_connection_values(connection, MHD_GET_ARGUMENT_KIND, get_query_callback, &data) < 0 || data.error) {
 		debug(LOG_DEBUG, "Unable to get query string - error 503");
-		/* no mem */
+		/* not enough memory */
 		return send_error(connection, 503);
```
lynxis, line 653: We should change this 503.
IMHO: the user is accessing a website with a long long query string (not sure what the spec says about the maximum length), but in the of redirection we should better go for dropping the whole query string.

mwarning: Agreed, we might trigger some unindentent action :D

mwarning: Hm, but that does not sound trivial. Maybe I should just drop that commit.

src/http_microhttpd.c
```diff
@@ -121,27 +121,33 @@ static int do_binauth(struct MHD_Connection *connection, const char *binauth, t_
...
+	struct get_query_data *data = cls;
+	const char separator = data->size ? '&' : '?';
+	const char *format = value ? "%c%s=%s" : "%c%s";
+	const int left = data->capacity - data->size;
+
+	if (key != NULL && kind == MHD_GET_ARGUMENT_KIND) {
+		// append '?foo=bar', '&foo=bar', '?foo', '&foo'
+		int n = snprintf(&data->query[data->size], left, format, separator, key, value);
+		if (n >= left) {
+			data->query[data->size] = '\0';
```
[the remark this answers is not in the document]
lynxis: The man page says snprintf() will write size including null termination. also my test case works that way.

src/http_microhttpd.c
lynxis: This is a different position than your first commit. You can leave your first commit in and I'll do a different patch based on top of it.

src/http_microhttpd.c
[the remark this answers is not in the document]
mwarning: You are right, but I meant another case. E.g.:

1. snprintf tries to append "foo=12345" to a buffer of 6 bytes
2. snprintf will write "foo=1\0" and return 9
3. the null assignment in line 142 will then discard the whole key/value pair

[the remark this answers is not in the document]
lynxis: Ack, keep it as it is.

mwarning:
@lynxis merged. Feel free to cleanup later on. :-)

PR #647, opened 2025-01-25
add github runner workflow for building debian packages for arm64 and amd64 architectures

sassanh:


mwarning:
hi @sassanh, thank you for the contribution. I do not have experience with gitlab CIs in general. Can you tell me were those debian packaghes will be found once the pipeline has run.

sassanh:
@mwarning The artifacts of the action are available in the action page, for example here:
https://github.com/sassanh/nodogsplash/actions/runs/12964179347
If you scroll down you should see these:
image
(scrolling down is a bit tricky in github actions page, you should scroll out of the canvas or it will be taken as zoom in/out for the canvas)

I'm working on a new commit to make it automatic to do releases by pushing tags like `v5.0.3` (tags starting with the letter `v`) and it will automatically attach the artifacts to the release.

sassanh:
Sample action for the release: https://github.com/sassanh/nodogsplash/actions/runs/12964307711
Sample release: https://github.com/sassanh/nodogsplash/releases/tag/v0.0.1

sassanh:
@mwarning 
I noticed packages built in ubuntu 2024.04, don't install on older debian versions so I changed it to create both 2022.04 and 2024.04 for each of arm64 and amd64 architectures. So there are 4 packages generated and attached to the release in total like this: https://github.com/sassanh/nodogsplash/releases/tag/v0.0.4

After this is merged, we need to do the releases by creating a tag named `vx.y.z` when `x.y.z` is the desired version and push the tag, it will automatically create packages, do the release and attach the packages as assets to the release.

Other than the automation, another advantage is people trust this automatically generated packages as their build process is transparent for public.

mwarning:
Please be aware that Nodogsplash is not actively maintained anymore. Any involvement is appreciated.

sassanh:
@mwarning Yes, I am aware of that. But considering people are still using it, I hope this commit makes it more accessible.

mwarning:
@stokito thank you for your feeback! I will apply your suggestions after the merge and @sassanh can then make any other changes in a separate MR.

But we don't know when there will be a new release. For that we need more contributions. If the contribution makes problems then we might remove it again if we do not feel maintaining it.

Issue #653, opened 2025-10-16
Can't be used on my padavan


lucktu:
When I use it on padavan (3.4.3.9-099, https://opt.cn2qq.com), I get such an error message, then exit.

`[3][Thu Oct 16 08:27:46 2025][6304](src/main.c:295) Could not create web server: Protocol not available`

The program I compiled with this method works normally on openwrt24(Mtk 7620) and the openwrt19 of docker(x86_64).

My compilation method is as follows
https://github.com/lucktu/compile/blob/main/.github/workflows/nodogsplash_2.yml

Where is the problem? 

Thank you!

lucktu:
<pre>[RT-N56U_B1 /etc/storage/bin]# cd /etc/storage/bin && nodogsplash -c nodogsplash.conf -f -s -d 3
[6][Mon Oct 20 08:10:24 2025][12134](src/main.c:371) Reading and validating configuration file nodogsplash.conf
[6][Mon Oct 20 08:10:24 2025][12134](src/conf.c:669) Reading configuration file 'nodogsplash.conf'
[7][Mon Oct 20 08:10:24 2025][12134](src/conf.c:705) Parsing option: GatewayInterface, arg: br0
[7][Mon Oct 20 08:10:24 2025][12134](src/conf.c:705) Parsing option: FirewallRuleSet, arg: authenticated-users {
[7][Mon Oct 20 08:10:24 2025][12134](src/conf.c:397) Parsing FirewallRuleSet authenticated-users
[7][Mon Oct 20 08:10:24 2025][12134](src/conf.c:437) p1 = [FirewallRule]; p2 = [allow all]
[7][Mon Oct 20 08:10:24 2025][12134](src/conf.c:579) Adding FirewallRule allow all port (null) to 0.0.0.0/0 to FirewallRuleset authenticated-users
[7][Mon Oct 20 08:10:24 2025][12134](src/conf.c:452) FirewallRuleSet authenticated-users parsed.
[7][Mon Oct 20 08:10:24 2025][12134](src/conf.c:705) Parsing option: FirewallRuleSet, arg: preauthenticated-users {
[7][Mon Oct 20 08:10:24 2025][12134](src/conf.c:397) Parsing FirewallRuleSet preauthenticated-users
[7][Mon Oct 20 08:10:24 2025][12134](src/conf.c:437) p1 = [FirewallRule]; p2 = [allow tcp port 53]
[...]

PR #655, opened 2025-11-03
Fix the error of refreshing the webpage, ipv6 and port multiplexing

lucktu:
lmq8267 Commits on Oct 23, 2025 https://github.com/lmq8267/nodogsplash/commit/45342cb533ef144352b2df014a0127946492bd73

### DIFF
--- a/src/conf.h
+++ b/src/conf.h
@@ -29,7 +29,7 @@
 #ifndef _NDS_CONF_H_
 #define _NDS_CONF_H_
 
-#define VERSION "5.0.3"
+#define VERSION "5.0.2"
 
 /*@{*/
 /** Defines */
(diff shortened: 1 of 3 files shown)


bluewavenet:
@lucktu 
Can you please give some explanation for why you need these changes. I don't see any descriptions anywhere.

For example why does MHD_OPTION_LISTENING_ADDRESS_REUSE have to be set now? Nothing has changed and it was not needed before. How is this related to "refreshing the web page"? I assume you are getting an authentication failure to give the splash page reload.

The ipv6 issue requires ip6tables as a dependency, not a good idea. Supporting  only "deprecated" iptables makes Nodogsplash a legacy application, useful on old, outdated, resource limited devices, so adding dependencies is not a good direction in which to move.

If you have a very specific use case for these changes, how will those changes effect mainstream use?

bluewavenet:
@lucktu 
Your Padavan issue was resolved:
https://github.com/nodogsplash/nodogsplash/issues/653#issuecomment-3441200801

src/conf.h
```diff
@@ -29,7 +29,7 @@
 #ifndef _NDS_CONF_H_
 #define _NDS_CONF_H_
 
-#define VERSION "5.0.2"
+#define VERSION "5.0.3"
```
mwarning: No need to change the version. It is done for releases.

bluewavenet:
@lucktu 
Nodogsplash is just fine for very old, resource limited routers running legacy operating systems.
I don't need to tell you where you can find support for ipv6 and per client traffic control....

mwarning:
@lucktu looks like someone is working on nftables support (see the other MRs). Let's see.

PR #656, opened 2025-11-11
tc.c: fix tc_detach_client

db4rne:
looks like the commands to execute tc in tc_detach_client were missing the actual tc call


lynxis:
@db4rne could you improve the commit message further? What are you fixing? What was wrong?

mwarning:
Let's merge after the commit message was improved a bit. Our commit messages never have been very good in NDS, but let's help others after us who want to understand what is going on. :D

lynxis:
It would be great if github would detect that this has been merged.
But was merged into main

PR #657, opened 2025-11-11
abstraction of firewall functions

db4rne:
intruducing an abstraction layer for firewall functions in order to prepare for nftables support

### DIFF
--- a/src/auth.c
+++ b/src/auth.c
@@ -36,6 +36,7 @@
 #include "debug.h"
 #include "auth.h"
 #include "fw_abstract.h"
+#include "fw_common.h"
 #include "client_list.h"
 #include "util.h"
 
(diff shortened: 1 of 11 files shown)


src/auth.c
```diff
@@ -312,8 +313,11 @@ auth_client_trust(const char *mac)
 
 	LOCK_CONFIG();
 
-	if (!add_to_trusted_mac_list(mac) && !iptables_trust_mac(mac)) {
-		rc = 0;
+	rc = fw_gops.trust_mac(mac);
+	if (rc) {
```
mwarning: The logic looks inverted here.
iptables_trust_mac() returns 0 on success. So it should if "if (rc == 0)" or "if (!rc)", no?

mwarning: Also probably in some other places.

lynxis:
Merged into main

PR #658, opened 2025-11-11
[RFC] support nftables firewall in nodogsplash

db4rne:
I think the basic stuff is working, I think there is some way to go to have ipv6 support, maybe someone can look at it and advise a bit?
traffic control is also not working, no idea yet if I will implement that.
traffic counter reporting should be easy though, especially since we have json-c as dependency now anyway

### DIFF
--- a/src/conf.h
+++ b/src/conf.h
@@ -29,6 +29,8 @@
 #ifndef _NDS_CONF_H_
 #define _NDS_CONF_H_
 
+#include "debug.h"
+
 #define VERSION "5.0.2"
 
 /*@{*/
@@ -89,6 +91,8 @@
 #define DEFAULT_EMPTY_AUTHENTICATED_USERS_POLICY "RETURN"
 #define DEFAULT_EMPTY_PREAUTHENTICATED_USERS_POLICY "REJECT"
 #define DEFAULT_IP6 0
+#define DEFAULT_USE_NFTABLES 0
+#define DEFAULT_NFTABLE_NAME "nodogsplash"
 /*@}*/
 
 /**
@@ -185,6 +189,8 @@ typedef struct {
 	char *binauth;			/**< @brief external authentication program */
 	char *preauth;			/**< @brief external preauthentication program */
 	char *statefile;		/**< @brief external preauthentication program */
+	bool use_nftables;
+	char *nftable_name;
 } s_config;
 
 /** @brief Get the current gateway configuration */
--- a/src/fw_nftables.c
+++ b/src/fw_nftables.c
@@ -1,0 +1,649 @@
+/********************************************************************\
+ * This program is free software; you can redistribute it and/or    *
+ * modify it under the terms of the GNU General Public License as   *
+ * published by the Free Software Foundation; either version 2 of   *
+ * the License, or (at your option) any later version.              *
+ *                                                                  *
+ * This program is distributed in the hope that it will be useful,  *
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of   *
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the    *
+ * GNU General Public License for more details.                     *
+ *                                                                  *
+ * You should have received a copy of the GNU General Public License*
+ * along with this program; if not, contact:                        *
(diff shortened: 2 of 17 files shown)


bluewavenet:
@mwarning 

>  I think they use libjson-c to store the state of NDS on restart 

Storing/restoring the state, including authenticated clients is a good thing. But using another dependency to first json encode then decode seems to be there only to say the data is in json format. A flat file or series of files would be very much more efficient. If the file is to be read by an admin or similar, then an optional separate package to massage the data into nice tables or even json format would be a better idea for this use case.

_**More important though, my two points above should be considered.**_

**Currently, if you have an ipv6 Internet feed, then Nodogsplash is broken.**

 **If a third party nftables firewall is present (as you will have for ipv4 routing), Nodogsplash is also probably broken.**

mwarning:
@bluewavenet the json-c dependency is not a topic for this MR. Let's focus on the nftables topic.

bluewavenet:
@mwarning 

> Let's focus on the nftables topic.

Indeed. My comments on that subject were an aside.
However the two points I raised earlier with respect to nftables migration are very important, ie address family and chain priority.

db4rne:
I updated the PR. The basic support is working and tested on OpenWRT. We are already using it on a few hundred devices and have no issues so far. of course, it's possible I overlooked something. I will add commits for traffic control and mss clamping now. I would be very glad if someone would take the time to review these changes and come back with some feedback.

for now, IPv6 is still out of scope, but once this is merged I'm planning on upstreaming support for IPv6 as well.

src/fw_nftables.h
```diff
@@ -0,0 +1,66 @@
...
+ * along with this program; if not, contact:                        *
+ *                                                                  *
+ * Free Software Foundation           Voice:  +1-617-542-5942       *
+ * 59 Temple Place - Suite 330        Fax:    +1-617-542-2652       *
+ * Boston, MA  02111-1307,  USA       gnu@gnu.org                   *
+ *                                                                  *
+\********************************************************************/
+
+/** @file fw_iptables.h
+    @brief Firewall iptables functions
```
lynxis: the copyright is wrong.

src/fw_nftables.c
```diff
@@ -0,0 +1,795 @@
...
+
+	return rc;
+}
+
+int
+_nftables_remove_mac_from_list(const char list[], const char mac[])
+{
+	int rc = 0;
+
+	// put mac on list
```
lynxis: wrong comment. but also you can remove it completely.
See lines above.

src/fw_nftables.h
```diff
@@ -0,0 +1,66 @@
...
+int nftables_fw_authenticate(t_client *client);
+int nftables_fw_deauthenticate(t_client *client);
+
+/** @brief Return the total download usage in bytes */
+unsigned long long int nftables_fw_total_download();
+
+/** @brief Return the total upload usage in bytes */
+unsigned long long int nftables_fw_total_upload();
+
+/** @brief All counters in the client list */
```
lynxis: Brief is here misleading. Correct would be `Update all counters...`

lynxis:
@db4rne I'll do another look onto it. Could you take a look onto the comments? They are mixed between `//` and `/* */`. I would stick to one and use `/* */`.
Additional some of the comments are superfluous and some can be removed without removing any information.

db4rne:
> @db4rne I'll do another look onto it. Could you take a look onto the comments? They are mixed between `//` and `/* */`. I would stick to one and use `/* */`. Additional some of the comments are superfluous and some can be removed without removing any information.

should be done now, I didn't touch the mixed styles in all the other files though

src/conf.h
```diff
@@ -89,6 +91,8 @@
 #define DEFAULT_EMPTY_AUTHENTICATED_USERS_POLICY "RETURN"
 #define DEFAULT_EMPTY_PREAUTHENTICATED_USERS_POLICY "REJECT"
 #define DEFAULT_IP6 0
+#define DEFAULT_USE_NFTABLES 0
```
mwarning, line 94: Why not?
```
#define DEFAULT_USE_NFTABLES false
```

PR #661, opened 2025-11-27
refactor get_client_mac

db4rne:
the get_client_mac function was a bit slow, so I extended it a bit to use /proc/net/arp if possible. since I'm not sure how this will work in non-linux environments, the old behaviour is the (automatic) fallback


db4rne:
I added tests now and fixed the function, would love to get this merged now

src/http_microhttpd.c
```diff
@@ -211,44 +211,7 @@ static int is_splashpage(const char *host, const char *url)
...
-	if ((len + 2) > sizeof(ip)) {
-		return -1;
-	}
-
-	// Extend search string by one space
-	memcpy(ip, req_ip, len);
-	ip[len] = ' ';
-	ip[len+1] = '\0';
 
-	stream = popen("ip neigh show", "r");
```
mwarning, line 234: I think the orignal idea of using "ip neigh show" was to support IPv6 as well. Not that IPv6 support has ever been completed in NDS... :D

Just something I find worth to mention from my side.

db4rne: hm, you are right, thats not good. Ipv6 should be supported. I will close this PR and open a new one, using netlink

PR #665, opened 2026-04-07
state_file: remove client ID from export/import

db4rne:
Let client_list_add_client() assign IDs automatically to prevent duplicate IDs after state file restore. Also fix fw_connection_state to use json_object_get_int instead of json_object_get_int64."

### DIFF
--- a/src/client_list.c
+++ b/src/client_list.c
@@ -91,9 +91,6 @@ client_list_flush(void)
 	t_client *client = firstclient;
 	t_client *prev = NULL;
 
-	if (client_count == 0)
-		return;
-
 	while (client != NULL) {
 		prev = client;
 		client = client->next;
--- a/src/client_list.c
+++ b/src/client_list.c
@@ -44,7 +44,7 @@
 
 /** Client counter */
 static int client_count = 0;
-static int client_id = 1;
+int client_id = 1;
 
 /** Global mutex to protect access to the client list */
 pthread_mutex_t client_list_mutex = PTHREAD_MUTEX_INITIALIZER;
--- a/src/client_list.h
+++ b/src/client_list.h
@@ -109,5 +109,6 @@ void client_list_flush(void);
 } while (0)
 
 extern pthread_mutex_t client_list_mutex;
+extern int client_id;
 
 #endif /* _NDS_CLIENT_LIST_H_ */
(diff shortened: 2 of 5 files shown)


db4rne:
I forgot the commits for the tests as well, it is complete now and ready for review

mwarning:
I am not familiar with the state file feature. So got ahead :D

db4rne:
@mwarning at the moment (without this fix), using the state file feature will result in having duplicate client IDs, and therefor clients that won't be able to connect.
is there something I can do to get this pulled into main?

mwarning:
@db4rne Maybe you can ask @lynxis for a review. If he does not have time, then I am fine with just merging it. :)

src/state_file.c
```diff
@@ -186,8 +178,8 @@ state_file_import_client(json_object *json_client)
 	JSON_GET_FIELD(client->counters.outgoing, err, counters, "outgoing", json_type_int, json_object_get_uint64);
 	JSON_GET_FIELD(client->counters.last_updated, err, counters, "last_updated", json_type_int, json_object_get_uint64);
 
-	unsigned int fw_connection_state = -1;
```
lynxis, line 189: Move this (both the get_int and explicit set it to a different value into an own commit

src/client_list.h
```diff
@@ -110,4 +110,8 @@ void client_list_flush(void);
 
 extern pthread_mutex_t client_list_mutex;
 
+#ifdef __NDS_UNIT_TEST
+int get_client_id();
```
lynxis: IMHO: to keep the code cleaner, you can remove the define NDS_UNIT_TEST.

Or as alternative make the client_id public accesable.

src/client_list.c
```diff
@@ -98,9 +98,6 @@ client_list_flush(void)
 	t_client *client = firstclient;
 	t_client *prev = NULL;
 
-	if (client_count == 0)
```
lynxis, line 94: A better solution would be ensuring the client_count is increased when importing the clients.
You can still remove this check, if you like. But at least keep the client_count in sync.

lynxis (changes requested):

src/client_list.c
db4rne: The commit message was wrong. client_count is already increased when importing the JSONs. I fixed the commit message

src/client_list.h
db4rne: done, I made it global

src/state_file.c
db4rne: done

PR #666, opened 2026-04-07
Fix/long config lines

db4rne:
at the moment, config lines longer than 4096 bytes will just be cut off, most likely preventing NDS from starting. this patchset will read any length of config line.
it also does some reordering of the tests, but it was not useful IMHO to put these in another PR, I hope thats ok.


mwarning:
Looks good to me code wise.

@db4rne  @lynxis to be want to use Signed-off-by in commit messages and a "<title>: <description>" commit description header structure?

PR #667, opened 2026-04-07
Feature/netlink neighbour discovery

db4rne:


src/util.c
```diff
@@ -164,6 +167,92 @@ int execute_ret(char* msg, int msg_len, const char fmt[], ...)
...
+	req.ndm.ndm_family  = family;
+
+	if (send(sock, &req, req.nlh.nlmsg_len, 0) < 0) {
+		debug(LOG_ERR, "get_client_mac: netlink send failed: %s", strerror(errno));
+		close(sock);
+		return 1;
+	}
+
+	char buf[4096];
+	int found = 0;
```
mwarning, line 215: We could use a bool here.

mwarning:
Netlink is probably faster compared to the use of shell commands.
@db4rne do you have an idea what speedup this has compared to the old code?

db4rne:
I did not test the speedup, the only reason I even went to improve the get_client_mac function was because it gets called for every client created. in a normal setup this is not really a concern, but if we have a big state file with X clients, this function gets called X times, so it resulted in a delay when importing big client lists.

db4rne:
forgot to say this, but I will try to add a testcase importing a lot of clients and compare the results. I'm also not sure what to do about the BSDs in this case, I'm not sure they are even still supported by NDS, can you comment on that?

mwarning:
We do not support BSD. It would be nice to have, but in that case we can pull the code back from the git history.

PR #668, opened 2026-07-27
state-file: switch from json-c to jansson

db4rne:
libnftables depends on libjansson for JSON creation and parsing. In order to save some memory, we should switch to jansson as well


mwarning:
For more context: firewall4 depends on nftables-json which dependes on libjansson

The MR looks good to me from a first look perspective.

PR #669, opened 2026-09-07
http_microhttpd: check URL size before copying

db4rne:
in the libmicrohttpd_cb, the user controlled _url gets passed to buffer_path_simplify(). _url is typically up to 32k Bytes big, while the url buffer is initialized with PATH_MAX. Depending on the system, this is 4K Bytes or 16k Bytes, but always smaller than the maximum _url buffer.
To prevent this, the _url size gets checked against PATH_MAX before sanitization, resulting in a HTTP error 414 if it's too long.


src/http_microhttpd.c
```diff
@@ -316,6 +316,12 @@ libmicrohttpd_cb(void *cls,
 	char url[PATH_MAX] = { 0 };
 	int rc = 0;
 
+
+	if (sizeof(_url) >= PATH_MAX) {
```
mwarning, line 320: You probably do not want to use sizeof here. ;-)

db4rne: the buffer for _url comes from libmicrohttpd, and it's bigger than our url buffer. so as far as I understand, we can use sizeof here without a problem.

mwarning: But sizeof(_url) will be always 8 on 64bit systems. It is the size of a pointer.

db4rne: oh, my... yes, you are right. I will fix that.

mwarning: I am glad that I could contribute a bit.

db4rne:
I updated the fix, please have another look, thank you!

